PR #17, opened 2021-08-03
Add scaling figure and section explaining scaling in README

ulupo:


README.rst
```diff
@@ -39,6 +39,16 @@ License
...
 If you need a different distribution license, please contact the `L2F team`_.
 
+Parallel persistent homology backend
+====================================
+
+Computing persistence barcodes of large datasets and in high homology degrees is challenging even on modern hardware. ``giotto-ph``'s persistent homology backend
+is able to distribute the key stages of the computation (namely, search for apparent pairs and coboundary matrix reduction) across an arbitrary number of available CPU threads.
+The scaling is generally quite favourable as shown in the following figure (for more details, see our paper linked below):
+
+.. image:: https://raw.githubusercontent.com/giotto-ai/giotto-ph/readme_speedup_figure/docs/images/multithreading_speedup.svg
```
ulupo, line 49:
```suggestion
.. image:: https://raw.githubusercontent.com/giotto-ai/giotto-ph/main/docs/images/multithreading_speedup.svg
```
To be changed after review

MonkeyBreaker, line 49: I would have a preference for centering the image, what do you think ? (https://stackoverflow.com/a/67479088)

MonkeyBreaker, line 49: I don't know if it would be better to remove sphere and dragon just for this specific plot ...
Just to sell us a bit more xD

ulupo: I tried placing the centering directive, but it doesn't work.  Apparently it is a known issue with GitHub: https://github.com/github/markup/issues/163

MonkeyBreaker: Oh, I did not know.
Apparently in the last answer from the issue, you could hack this by providing directly the html code.
See https://github.com/github/markup/issues/163#issuecomment-846676443

ulupo: Heh, I'd prefer no cherry-picking and no deviations from the paper, but maybve we can defend ourselves by saying something like:

> On challenging datasets, the scaling is quite favourable as shown in the following figure (for more details, see our paper linked below):

?

MonkeyBreaker: I am ok with your suggestion. It was my Marketing blood who tried to suggest evil things :)

ulupo: Yep, let me try this approach.

ulupo: I can't make it work!

ulupo: Done!

ulupo: @MonkeyBreaker FYI, the reason for not using relative paths is that, when we were doing this in ``giotto-tda``, the image was not rendered in the PyPI readme

MonkeyBreaker: Oh ! I was not aware, good to know !

ulupo: Reverted to standard RST image directives

ulupo: Done. Ready as far as I am concerned. Let me know :)

MonkeyBreaker: You can merge

PR #23, opened 2021-08-16
Add cocycles feature

MonkeyBreaker:
This PR allows users to retrieve cocycles from the Persistent Homology computation.

* [x] : C++ backend compute cocycles during computation
* [x] : Update bindings to enable retrieving cocycles
* [x] : Update Python interface to support retrieving `cocycles` with argument `do_cocycles`
    * [ ] : Should we use a different name than `do_cocycles` ?
* [ ] : Add test for verify that the feature works as expected
    * [ ] : Add test when cocycles is produced by a persistence pair
    * [ ] : Add test when cocycles is produced by an essential pair
* [ ] : Add a notebook with an example of using `cocycles` with maybe some interaction with a plot for a nice visualization.


gph/bindings/ripser_bindings.cpp
```diff
@@ -15,85 +15,84 @@
...
-    // ImportError: generic_type: type "ripserResults" is already registered!
-    // When same python module imports gtda_ripser and gtda_ripser_coeff
-    py::class_<ripserResults>(m, "ripserResults", py::module_local())
-        .def_readwrite("births_and_deaths_by_dim",
-                       &ripserResults::births_and_deaths_by_dim)
-        .def_readwrite("num_edges", &ripserResults::num_edges);
+  // Because `ripser` could have two different modules after compilation
+  // It's necessary to add `py::module_local()` to prevent following issue:
+  // ImportError: generic_type: type "ripserResults" is already registered!
+  // When same python module imports gtda_ripser and gtda_ripser_coeff
```
ulupo, line 29: Are these comments still relevant?  Because they refer to `gtda_ripser` and `gtda_ripser_coeff`

gph/bindings/ripser_bindings.cpp
```diff
@@ -15,85 +15,84 @@
...
+        compressed_lower_distance_matrix dist =
+            compressed_lower_distance_matrix(
+                compressed_upper_distance_matrix(std::move(distances)));
 
-            // TODO: This seems like a dummy parameter at the moment
-            float ratio = 1.0;
-            int num_edges = 0;
+        // TODO: This seems like a dummy parameter at the moment
+        float ratio = 1.0;
+        int num_edges = 0;
```
ulupo, line 49: Why not remote `ratio` if we don't use it?  We could still create an issue about it if we are not sure whether it should be implemented.

ulupo, line 49: Do we return `num_edges`?  It is such a simple operation and I'm not sure we want to waste computational time for it especially if it's not returned.

gph/bindings/ripser_bindings.cpp
```diff
@@ -15,85 +15,84 @@
...
-                }
-            }
-            ripserResults res;
-            r.copy_results(res);
-            res.num_edges = num_edges;
-            return res;
-        },
-        "I"_a, "J"_a, "V"_a, "NEdges"_a, "N"_a, "modulus"_a, "dim_max"_a,
-        "threshold"_a, "num_threads"_a, "ripser sparse distance matrix");
+        // TODO: This seems like a dummy parameter at the moment
```
ulupo, line 76: As above for `ratio` and `num_edges`

ulupo (changes requested):
Just a first quick look at `ripser_bindings.cpp`

gph/bindings/ripser_bindings.cpp
MonkeyBreaker: They are still relevant, but as you pointed out they are not up-to-date

gph/python/ripser_interface.py
```diff
@@ -347,6 +351,10 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
         homology computation. When passing ``-1``, it will try to use the
         maximal number of threads available on the host machine.
 
+    do_cocycles: bool, optional, default False
```
ulupo, line 354:
```suggestion
    do_cocycles: bool, optional, default: ``False``
```

gph/bindings/ripser_bindings.cpp
MonkeyBreaker: We should remove it entirely, the operation done by `ratio` can easily be implemented in Python after the computation of PH.

MonkeyBreaker: But I don't know if we do it in this PR, or we should create a PR for more misc tasks that needs to be performed ?

MonkeyBreaker: We return it. Even in the C++ code it is not really costly. In my opinion we can keep it because it doesn't have a significant impact on run-time. But, in the Python interface we don't use it for the moment.

ulupo: I'd say this should be moved to an issue/other PR.

MonkeyBreaker: Yes, we should integrate it just before releasing 0.2 IMO

ulupo: I guess I would make the same point as you made for `ratio`.  I would slightly prefer if the bindings were really minimalistic and there was no computation done in them.

ulupo:
> Should we use a different name than do_cocycles ?
Even if longer, I would slightly prefer `return_cocycles`.  Reason: one can "do" a lot of computation without returning anything.

MonkeyBreaker:
> > Should we use a different name than do_cocycles ?
> 
> Even if longer, I would slightly prefer `return_cocycles`. Reason: one can "do" a lot of computation without returning anything.

Sure, should it be an argument of function `ripser_parallel` or should we pass it via `kwargs` ?

ulupo:
> Sure, should it be an argument of function ripser_parallel or should we pass it via kwargs ?

Are you asking if it should be positional or keyword-based with a default?  I would strongly support keyword-based with a default of ``False``, as you have already done/

gph/bindings/ripser_bindings.cpp
MonkeyBreaker: You are maybe right. Then lets add it to the list of *minor* changes to add for release `0.2`

gph/bindings/ripser_bindings.cpp
MonkeyBreaker: I added a note directly into the project 0.2, we could maybe resolve the conversations ?

ulupo: Great!  So we address these in another PR, and I resolve here.

gph/src/ripser.h
```diff
@@ -538,7 +547,22 @@ typedef struct {
...
+      itself an array of unrolled information about the cocycle For dimension 1,
+      for example, the zeroeth element of the array contains [ccl0_simplex0_idx0
+      ccl0_simplex0_idx1 ccl0_simplex0_val, ccl0_simplex1_idx0
+      ccl0_simplex1_idx1 ccl0_simplex1_val, ... ccl0_simplexk_idx0
+      ccl0_simplexk_idx1 ccl0_simplexk_val] for a cocycle representing the first
+      persistence point, which has k simplices with nonzero values in the
+      representative cocycle
+    */
+    std::vector<cocycles_t> cocycles_by_dim;
     /* The second variable is the number of edges that were added during the
```
ulupo, line 566:
```suggestion
    /* The third variable is the number of edges that were added during the
```

ulupo (changes requested):

gph/python/ripser_interface.py
```diff
@@ -356,6 +364,14 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             than maxdim. Each diagram is an ndarray of size (n_pairs, 2)
             with the first column representing the birth time and the
             second column representing the death time of each pair.
+        'cocycles': list (size maxdim) of list of ndarray, optional
+            For each dimension less than 'maxdim' a list of representative
+            cocycles. Each representative cocycle in dimension 'd' is
+            represented as a ndarray of '(k,d+1)' elements. Each non zero value
```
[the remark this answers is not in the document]
MonkeyBreaker: Good catch

MonkeyBreaker:
@ulupo updated !

gph/bindings/ripser_bindings.cpp
```diff
@@ -39,15 +39,13 @@ to_numpy_barcodes(std::vector<barcodes_t> barcodes)
...
 #else
-PYBIND11_MODULE(gph_ripser, m)
-{
+PYBIND11_MODULE(gph_ripser, m) {
 #endif
 
-    using namespace pybind11::literals;
-    m.doc() = "Ripser python interface";
+  using namespace pybind11::literals;
+  m.doc() = "Ripser python interface";
```
ulupo, line 48: Is this indentation intentionally 2 spaces instead of 4?

gph/bindings/ripser_bindings.cpp
```diff
@@ -63,14 +61,15 @@ PYBIND11_MODULE(gph_ripser, m)
...
+        .def_readwrite("cocycles_by_dim", &ripserResults::cocycles_by_dim)
         .def_readwrite("flag_persistence_generators_by_dim",
                        &ripserResults::flag_persistence_generators);
 
     m.def(
         "rips_dm",
         [](py::array_t<value_t>& D, py::array_t<value_t>& diag, int modulus,
            int dim_max, float threshold, int num_threads,
-           bool return_generators) {
+           bool return_generators, const bool do_cocycles) {
```
ulupo, line 72: Indentation: 3 spaces?

ulupo:
Thanks a lot @michiexile!  Do you think there is a bug in the cocycles implementation in this PR, or in the necessary post-processing to produce the graphics?

michiexile:
> Thanks a lot @michiexile! Do you think there is a bug in the cocycles implementation in this PR, or in the necessary post-processing to produce the graphics?

I honestly don't know, and I haven't had time to dig into it. The post-processing, we've been using relatively successfully with `ripser` and almost identical code. That's why I was talking about `dperm2all` above - `ripser` does some permutations to speed things up, and "reports back" by returning a permuted distance matrix. If you too are doing similar permutations, they would need to be reported back for my script to do the right thing.

ulupo:
I see.  Since `giotto-ph` does not include greedy permutations, I'd assume things should work just as they would in `ripser.py` when greedy permutations are not used there.  Would this be a simple test for you to perform?

MonkeyBreaker:
Hi @michiexile,

Thank you so much for helping us with this. I looked at your notebook (thank you for sharing it with us ;)), at I compared the results with the cocycles obtained with `ripser.py`.
It seems I messed up the representation (signed/unsigned) in the back-end, sorry for that.
I "fixed" this, and I am now able to obtain the same cocycles, the rainbow looks really nice !

Regarding:
> That's why I was talking about dperm2all above - ripser does some permutations to speed things up, and "reports back" by returning a permuted distance matrix. If you too are doing similar permutations, they would need to be reported back for my script to do the right thing.

If I am not mistaken, the permutations in `ripser.py` only happens if `n_perm` parameter is set. Otherwise `dperm2all` will contain the distance matrix used during the computation. Which in our case we decide to let the user compute it separately if needed (@ulupo maybe we should add it as output ?).

Again thank you so much to help us, please let me know if I can do anything here ;)

michiexile:
> It seems I messed up the representation (signed/unsigned) in the back-end, sorry for that.
> I "fixed" this, and I am now able to obtain the same cocycles, the rainbow looks really nice !

Great! Thank you!!

> If I am not mistaken, the permutations in ripser.py only happens if n_perm parameter is set. Otherwise dperm2all will contain the distance matrix used during the computation. Which in our case we decide to let the user compute it separately if needed (@ulupo maybe we should add it as output ?).

It's not clear to me what you might want to do with an explicit cocycle that does _not_ involve the explicit boundary matrix in one form or another - in my example code, we use the distance matrix to create the relevant boundary matrix. Maybe add them (as many of "distance matrix" and "boundary matrix" as you can provide) as optional outputs?

MonkeyBreaker:
Hey everyone,

Sorry I was unexpectedly busy this last week ...
But now I will try to push things forward with your help 😄 .

About returning the computed distance Matrix, we did it before and we remove it because of memory and performance implications if I remember right.
I don't have a really strong opinion on the matter, I don't think that adding an additional parameter for returning the computed the distance matrix is a good idea, make the interface more heavy. But, I am not sure neither if we should always return when cocycles are enable the distance matrix (prevent memory duplication if the user provided already a distance matrix to ripser interface).

What are your thoughts on this (@ulupo and @michiexile), if you think it makes sense to always return it when computing cocycles makes sense, then sure, lets do it.

Also, another point, @michiexile would you have an idea of an example of cocycles related to an essential pair (non-finite) ? Is that even possible ?

Thank you again for the help you provide us with :) !

michiexile:
> About returning the computed distance Matrix, we did it before and we remove it because of memory and performance implications if I remember right.
> I don't have a really strong opinion on the matter, I don't think that adding an additional parameter for returning the computed the distance matrix is a good idea, make the interface more heavy. But, I am not sure neither if we should always return when cocycles are enable the distance matrix (prevent memory duplication if the user provided already a distance matrix to ripser interface).
> 
> 

Shouldn't it be _really_ easy to detect whether the user is providing a distance matrix or not, and only produce the distance matrix when the user asks for cocycles _and_ has not provided a distance matrix themselves?
> What are your thoughts on this ***@***.*** <https://github.com/ulupo> and @michiexile <https://github.com/michiexile>), if you think it makes sense to always return it when computing cocycles makes sense, then sure, lets do it.
> 
> 

I have a hard time thinking of a situation where I would want the cocycles explicitly, but have no need for the distance matrix (or more specifically, no need for the VR-complex coboundary map). If you're willing to build the coboundary map for me, I can probably survive without a distance matrix pretty often.
> Also, another point, @michiexile <https://github.com/michiexile> would you have an idea of an example of cocycles related to an essential pair (non-finite) ? Is that even possible ?
> 
Ummmm... I would pretty much _always_ work with essential pairs. Because non-essential pairs no longer produce cocycles at the radius you are looking at (because they have both a birth and a death time smaller than the "current" time)

So the typical way I would be computing with circular coordinates goes something like this:

1. (optional) Compute a barcode to infinity, or to some high parameter value. Look for a parameter value during which a feature I am interested in studying will be essential.
2. Compute cocycle representatives at this parameter value, chosen specifically to make some of the pairs essential.
3. Compute the coboundary matrix at this parameter value.
4. Use cocycle and coboundary to compute circular coordinate.

Key to this workflow is being able to pick a parameter value to force some cocycles to become essential with respect to the (truncated) filtration we are looking at. Because if they are not, they will not actually be cocycles, and the computation in step 4 will fail.

I hope this helps describe the usecase I'm interested in?

All the best,
--
Mikael Vejdemo-Johansson
***@***.*** ***@***.***>

MonkeyBreaker:
> Shouldn't it be _really_ easy to detect whether the user is providing a distance matrix or not, and only produce the distance matrix when the user asks for cocycles _and_ has not provided a distance matrix themselves?

Sorry if I made it sound like it could be difficult. My concern here is if user provides an already computed distance matrix and we return it, that we don't duplicate memory but just forward (move) the memory. Per example, with a dataset of 50k points, it took on my machine like 10GB, if we are not careful, it could take 20GB when returning the distance matrix, because Python copies to return it.

I will try this week to prepare something here and check if we don't have this issue of memory duplication.

> I have a hard time thinking of a situation where I would want the cocycles explicitly, but have no need for the distance matrix (or more specifically, no need for the VR-complex coboundary map). If you're willing to build the coboundary map for me, I can probably survive without a distance matrix pretty often.

I think the coboundary map could be interesting, but would require some heavy changes in c++ part. I would for the moment avoid this but if this might be useful for additional use cases, then, it would be part of another PR.

> Ummmm... I would pretty much _always_ work with essential pairs. Because non-essential pairs no longer produce cocycles at the radius you are looking at (because they have both a birth and a death time smaller than the "current" time)

Right, sorry I was confused, I don't have a background in mathematics, without @ulupo I am lost sometimes.

> I hope this helps describe the usecase I'm interested in?

Definitely, I need to prepare a reproducible test where I always use the same input dataset and compute twice cocycles with a different threshold on each run. But, as Umberto already discussed with me, we would need to have access to the coboundary matrix, that, currently is only present in the c++ part...

Anyway, don't worry to much on this, I will still think about it.
On the meantime, I will try this week to provide the computed distance matrix as an output, do you have a suggestion of output parameter name ? Like `ripser.py` is calling it `dperm2all`, but in our case we don't yet perform any computations.

Also, concerning your notebook, do you need additional features from our side ?

Best,
Julián

ulupo:
I'm on vacation now and can only reply very briefly but this is just to say that returning the coboundary matrix may not be a terrible idea if it saves the user the extra effort.  @MonkeyBreaker I already coded this for the unit tests that should be included in this PR and I included it in a notebook I shared with you a while back.

@michiexile is there something the distance matrix would provide beyond the ability to build the coboundary matrix yourself?

Sorry if I'm missing something due to the rush.

MonkeyBreaker:
Hey,
Didn't have much time this week, and it doesn't seem I will have much next week.

> I'm on vacation now and can only reply very briefly but this is just to say that returning the coboundary matrix may not be a terrible idea if it saves the user the extra effort. @MonkeyBreaker I already coded this for the unit tests that should be included in this PR and I included it in a notebook I shared with you a while back.

I think that returning the coboundary matrix would be a good idea, but not for this PR (or if we compute it in Python). The reason is that one of the tricks user in `ripser` to accelerate computation and also consume the smallest amount of memory is to not store the coboundary matrix. But to generate it on the fly, this means that introducing this feature could require a lot of heavy changes in the backend. But maybe I am misunderstanding something here, lets have a chat once you are back from holidays ;) .

Otherwise, returning the distance matrix if computed, only for when returning cocycles, I am fine with this. I think we have 3 cases:
* No threshold applied from the user: We return the dense computed distance matrix.
* A threshold was applied by the user: We return a `coo_matrix((data, (row, col))` ? (or maybe a `csr_matrix` ?)
* When weights enable, do we return before weight computation or the one after weight computation ?

ulupo:
@MonkeyBreaker, thanks for the comments!  What I meant to say initially is that we could compute the coboundary matrix in pure Python, as I already partly did that work for those unit tests I mentioned.

But now I think that, while those unit tests should be there, it would be a bit ambitious to construct the coboundary matrix in full generality.  I only covered small homology dimensions in those tests, and only mod 2 coefficients.  I could do it in full generality by including further dependencies (such as PHAT or GUDHI), but we probably shouldn't go there just yet.

So I also now agree that we should return the distance matrix only if the user wants cocycles.  We should always return the distance matrix just before it is passed to the C++ backend.  So, in the case with weights, the distance matrix after reweighting.

michiexile:
I can't think of anything the distance matrix brings us that the coboundary matrix doesn't (and if the distance matrix is _really_ important, the user can and will probably compute it themselves for those other applications)

What I _do_ however need to be able to do is to carve out the coboundary matrix _at a specific threshold_. It is not enough for me to get the coboundary matrix at the end of the computation - for the circular coordinate computations to work, I need to be able to get the coboundary during a point where the cocycles I want to study are actually alive.

All the best,
--
Mikael Vejdemo-Johansson
***@***.*** ***@***.***>

PR #28, opened 2021-10-17
Fix issue #26 and remove treatment of 1D case

### LINKED ISSUE #26: Computation of enclosing radius is too slow
https://github.com/giotto-ai/giotto-ph/blob/1d0c628fe5e5c5c4f712bbc1fb9ddc3acc347f9d/gph/python/ripser_interface.py#L246 is too slow.   A faster solution is to do `np.min(np.max(dm, axis=1))`.

ulupo:
Fixes #26 as suggested there.

### DIFF
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -241,18 +241,10 @@ def _ideal_thresh(dm, thresh):
     if dm.shape[0] != dm.shape[1]:
         return thresh
 
-    # Compute an array containing the maximal value by column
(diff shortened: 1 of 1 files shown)


ulupo:
I think supporting 1D arrays would make sense if we allowed the user to pass condensed distance matrices, but I don't think this input is allowed by ripser_parallel for now.  So I think it's ok to not treat the 1D case at all here, since it is a in hidden function and so we can assume that the input was already validated by ripser_parallel.  What do you think?

PR #29, opened 2021-10-18
Flag persistence generators

ulupo:
This PR add the possibility to retrieve representative simplicies.

* [x] : Compute and retrieve finite representative for dimensions > 0
* [x] : Compute and retrieve essential representative for dimensions > 0
* [x] : Compute and retrieve finite representative for dimension == 0
* [x] : Compute and retrieve essential representative for dimension == 0
* [x] : Add support in the bindings for retrieving representative simplicies
* [x] : Add support in Python to retrieve and output representative simplicies
    * [x] : Document the output regarding representative simplicies
    * [x] : Throw exception if `collapser` and `representative simplicies` are enable at the same time
* [x] : Update documentation about representative simplicies
* [x] : Update example basic tutorial to show example of this feature


ulupo:
> (i for column and j for row)
We are playing with fire here :) I suggest we stick to `i` indexing rows and `j` columns as this is virtually universal in mathematics and Python.

ulupo:
During work on this PR #29, @MonkeyBreaker and I realized that the current core logic in the matrix reduction algorithm makes it so that we break the following promise to the user: that passing a dense matrix with a few `np.inf` will lead to exactly the same output as passing a sparse version where the `np.inf` entries are just missing entries.  This was never noticed at the level of barcodes because a "finite" bar with death equal to infinity looks just the same as an infinite bar with the same birth.  But flag generators don't lie: we now end up telling the user there are never any generators for essential classes if the matrix is dense, no matter what the matrix is or represents.

Because changing this requires potentially delicate changes to the reduction routine, we decided to discuss what to do about this separately, and open a separate PR if necessary after this one is merged.

ulupo:
In view of #36 , we need to update the imports in the two notebooks here.

MonkeyBreaker:
> In view of #36 , we need to update the imports in the two notebooks here.

Before doing the merge, we will need to rebase to `main` and then we will be able to merge, otherwise as you point out, we will have conflicts

gph/src/ripser.h
```diff
@@ -1163,6 +1229,52 @@ class ripser
...
+                            edge = new_cand;
+                        }
+                    }
+                }
+            }
+        }
+
+        return edge;
+    }
+
```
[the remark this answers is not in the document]
MonkeyBreaker: This was done on commit 0abb23b
I think we can resolve the issue, thank you Umbe for the suggestion of using a diamong input

MonkeyBreaker:
> Another note: res['gens'][1] and res['gens'][3] should always have length maxdim. Currently, they can be empty lists if there is no content. Instead, we should always return a list of length maxdim with empty ndarrays if necessary in some (or even all) positions.

It seems weird to me, the C++ core has the same logic as for barcodes. The only difference is that you have `maxdim-1` as size of res['gens'][1] and res['gens'][3] , because dimension 0 is not included.

I need to verify if this is not the case

ulupo:
> The only difference is that you have maxdim as size of res['gens'][1] and res['gens'][3] , because dimension 0 is not included.

I don't think this can be the case if `maxdim` means the same as in Python. Example: if `maxdim=1` (compute up to and including dimension 1) then the list has length 1, not 1-1=0.

gph/src/ripser.h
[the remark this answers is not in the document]
ulupo: Thanks for implementing it as a test!!

MonkeyBreaker:
> I don't think this can be the case if maxdim means the same as in Python. Example: if maxdim=1 (compute up to and including dimension 1) then the list has length 1, not 1-1=0.

Well when you compute `maxdim=1` you have dimension 0 and dimension 1 => 2-1 = 1, no ?

ulupo:
Oh I'm sorry @MonkeyBreaker , I thought you said `maxdim - 1` (because it used to say that in the docstrings), but you didn't! We said the same thing.

MonkeyBreaker:
> Oh I'm sorry @MonkeyBreaker , I thought you said maxdim - 1 (because it used to say that in the docstrings), but you didn't! We said the same thing, i.e. that res['gens'][1] and res['gens'][3] should have length maxdim.

Does it work as expected then ? Do we have an issue in the documentation ? (In the documentation we are right in that we return `maxdim` for [1] and [3].

ulupo:
> Does it work as expected then ? 

I think I was able to produce examples which returned empty lists when `maxdim=1`, but can't remember how.  Maybe `square_dm`?

MonkeyBreaker:
>  Maybe square_dm?

Just tried with `maxdim=1` works as expected

examples/flag_persistence_generators.ipynb
```diff
@@ -0,0 +1,168 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7e4bf17b-f6b4-444c-a344-38ff096c7f4a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Giotto-ph\n",
+    "from gph.python import ripser_parallel\n",
```
ulupo, line 33: The import here was not updated correctly, I wonder if something went wrong in resolving conflicts @MonkeyBreaker ?

ulupo, line 33:
```suggestion
    "from gph import ripser_parallel\n",
```

ulupo: I fixed it, and I suspect nothing went wrong. Merging!

MonkeyBreaker: :angry: I failed the merge, sorry for that

gph/python/test/test_ripser.py
```diff
@@ -189,3 +190,123 @@ def test_multithread_consistent(dm):
...
+        # Verify birth of dim-0 essential representative
+        # The birth is located on the diagonal
+        assert np.isclose(barcode[0], X[rp, rp])
+
+
+def test_gens_order_vertices_higher_dimension():
+    """This test verifies that function get_youngest_edge_simplex
+       returns the correct vertices. We are interested that in the computation
+       it will return the edge with the youngest vertices a the biggest diameter
+    """
```
ulupo, line 296:
```suggestion
    """This test verifies that function `get_youngest_edge_simplex` returns the
    correct vertices. It should return the edge with the largest diameter in
    the simplex and, if several are present with the same diameter, the oldest
    one in the reverse colexicographic order used to build the simplexwise
    refinement of the Vietoris-Rips filtration."""
```

PR #33, opened 2021-10-25
Remove explicit garbage collection step

### LINKED ISSUE #25: Garbage collection step in python interface is too runtime-costly
It seems to me that the price to pay for freeing up some memory in https://github.com/giotto-ai/giotto-ph/blob/1d0c628fe5e5c5c4f712bbc1fb9ddc3acc347f9d/gph/python/ripser_interface.py#L470-L471 is too high in terms of runtime when the input is small.  This will cause `ripser_parallel` to be too slow in applications with repeated calls on small inputs.

This step was introduced to address the fact that, for large input distance matrices, both a 2D square version and a compressed version will be created inside `ripser_parallel`.  Ultimately, I think a better way to address this issue will be to compute compressed distance matrices whenever possible, and to allow the user to pass compressed distance matrices.

ulupo:
Fixes #25


MonkeyBreaker:
Hi,

I had time to run some numbers ! I used the following script to measure the impact on small datasets:

```python
import time
import numpy as np
from scipy.spatial.distance import pdist, squareform

from gph.python import ripser_parallel as ripser

X = squareform(pdist(np.random.random((200, 3))))

timing = []
for _ in range(100):
    start_time = time.time()
    ret = ripser(X, metric='precomputed', maxdim=1, n_threads=1)
    timing.append(time.time() - start_time)

timing = np.array(timing)
print("--- Average {} seconds ---".format(np.mean(timing)))
```

I tested 3 different cases:

1. Before the PR
2. Removing the garbage collection step
3. Removing garbage collection and `del` step

| case | time [s] |
|-|-|
| original | 0.018 |
| without gc | 0.010 |
| without gc and del | 0.010 |

It seems that for small dataset, that PH computation takes less than 10ms, the cost of running the garbage collector doubles the runtime, this is not good and the PR make sense. `del` seems to not have any impact on runtime.

I then looked on the memory consumption when using as input `torus` dataset (point-cloud with 50000 points). Well to my big surprise, each of the runs consumed the exact same memory. I think that we kept this memory management step, when we originally were using the same computation as `ripser.py`.

IMO, we can also remove the `del` step that from my measurements does not add anything to the computation in terms of memory or runtime.

ulupo:
@MonkeyBreaker fantastic! Thanks a lot! So I'll remove the del and then hopefully we can close this.

Issue #24, opened 2021-10-04
Possibly faster get_simplex_vertices for 1-d simplices?


ulupo:
Currently,  `get_simplex_vertices` is implemented by iterating a call to `get_max_vertex`, and this in turn is implemented via a binary search.  But in the case of 1-dimensional simplices, which correspond to entries in the compressed distance matrix, a closed-form solution exists.  For upper triangular distance matrices in compressed format, it is explained e.g. in https://stackoverflow.com/a/36867493.

An idea could be to benchmark this and see if it gives any non-trivial speedup when the maximum homology dimension to compute is 1.

PR #38, opened 2021-11-01
[CPP] Specialized simplex vertices

ulupo:
A closed-form approach is implemented for 1D simplices as suggested in #24. It seems to hold up numerically for all dataset sizes of current interest.

Furthermore, the binary search is replaced in 2D by a dumb linear search starting from a "best guess" which is always within 1 of the true value (according to my current numerical tests).


gph/src/ripser.h
```diff
@@ -706,11 +729,19 @@ class ripser
...
             *out++ = n;
             idx -= binomial_coeff(n, k);
         }
+
+        double to_sqrt = 8 * idx + 1;
+        n = static_cast<index_t>(std::floor(1.5 + (0.5 * std::sqrt(to_sqrt)))) - 1;
+        *out++ = n;
+        idx -= binomial_coeff(n, 2);
+        *out++ = idx;
+        idx = 0;
```
MonkeyBreaker, line 743: This should be simplified to:

```cpp
*out = idx;
```

No need the increment and the `idx=0`

ulupo: The `idx=0` is just there to have consistent behaviour with the previous version, which always ended there. Mostly for debugging now, we can remove it at the end!

gph/src/ripser.h
```diff
@@ -691,9 +691,32 @@ class ripser
...
+                }
+            } else {
+                while (true) {
+                    ++guess;
+                    if (!pred(guess)) {
+                        --guess;
+                        break;
+                    }
+                }
+            }
```
ulupo, line 717: This routine is based on the observation that `guess` seems to always be within distance 1 of the true value.  Likely there is a closed-form formula with some tiny correction that we could work out, but I think we'd still need the left-and-right checks due to numerics that become more important for 2D simplices given the large index values.

gph/src/ripser.h
```diff
@@ -706,11 +729,19 @@ class ripser
...
         --n;
-        for (index_t k = dim + 1; k > 0; --k) {
+        for (index_t k = dim + 1; k > 2; --k) {
             n = get_max_vertex(idx, k, n);
             *out++ = n;
             idx -= binomial_coeff(n, k);
         }
+
+        double to_sqrt = 8 * idx + 1;
+        n = static_cast<index_t>(std::floor(1.5 + (0.5 * std::sqrt(to_sqrt)))) - 1;
```
ulupo, line 739: This formula is exact; numerics don't seem to be an issue but we should test this more rigorously.

gph/src/ripser.h
```diff
@@ -691,9 +691,32 @@ class ripser
...
         const index_t cnt = n - (k - 1);
+
         if (pred(n) || (cnt <= 0))
             return n;
-        return get_max(n, cnt, pred);
+
+        if (k != 3) {
+            return get_max(n, cnt, pred);
+        } else {
+            long double to_cbrt = 6 * idx;
```
ulupo, line 701: `long double` because indices are large here.

gph/src/ripser.h
```diff
@@ -691,9 +691,32 @@ class ripser
...
+            } else {
+                while (true) {
+                    ++guess;
+                    if (!pred(guess)) {
+                        --guess;
+                        break;
+                    }
+                }
+            }
+            return guess;
```
MonkeyBreaker, line 718: Suggestion:

```cpp
bool increment = pred(guess);

while (true) {
    if (increment) {
	++guess;
	if (!pred(guess)) {
	    --guess;
	    break;
	}
    } else {
	--guess;
        if (pred(guess))
            break;
    }
}

return guess;
```

Or 

```cpp
bool increment = pred(guess);

while (true) {
    if (increment) {
	++guess;
	if (!pred(guess)) {
	    break;
	}
    } else {
	--guess;
        if (pred(guess))
            break;
    }
}

return increment ? guess - 1 : guess;
```

gph/src/ripser.h
MonkeyBreaker: This should provide a measurable speed-up IMO

ulupo: We have to make sure `idx` is not used ever after calling this, I think there is an instance in the code where that is the case,

gph/src/ripser.h
MonkeyBreaker: `idx` is of type `index_t` which is of type `int64_t`.
Is there a reason that we convert to double ?
IMO, we could remain in as an integer, but If we are sure we do not overflow, we could go with `uint64_t`, but if it can overflow, we can let `index_t` and verify that `to_cbrt` is not `<0`.

gph/src/ripser.h
MonkeyBreaker: Do you have a link about the theory of this or is this common in your field ?

gph/src/ripser.h
ulupo: Don't these solutions hard-code the assumptions that the right answer is never 1 step away from the initial guess?

gph/src/ripser.h
MonkeyBreaker: I am not sure I agree. After the last `*out=idx` we do not need anymore `idx`. Inside of function `get_simplex_vertices` the `idx` value was passed by copy, it is not a reference. It will be discarded after we arrive at the end of the function.

gph/src/ripser.h
ulupo: We need to take cube roots of a huge number, so a floating-point type is necessary, right?

gph/src/ripser.h
MonkeyBreaker: No, we have first the `while(true)` loop and then we compute the `if` condition. We can increase/decrease multiple times until one of the break condition is true.

gph/src/ripser.h
ulupo: The cube root is "obvious", and even the leading 6 can be understood easily I think.  But for now I just have verification until very, very large numbers (I need to get my files again to tell you quite how large).

ulupo: In any case, this solution should always be correct as long as the initial guess is tested.

gph/src/ripser.h
ulupo: Ah, sorry, I mis-read it.  The first solution is certainly right.  What's going on in the second one and why should we prefer it?

gph/src/ripser.h
MonkeyBreaker: Yeah but the thing is that with double it seems that you can be less precise. See [link](https://ifnotnil.com/t/why-is-double-larger-than-int64/255).
Here also we are only doing a multiplication by an integer, this is a reason why I would keep it all integer, even if we can go a higher with double. Because for those big numbers, we will have less precision in double because we only work with the integer part.

gph/src/ripser.h
ulupo: > Inside of function get_simplex_vertices the idx value was passed by copy, it is not a reference.

Thanks for the clarification!  Yes then you are 100% right.

gph/src/ripser.h
MonkeyBreaker: It is not better IMO. It just keeps the core of the loop almost the same. And the `--guess` performed just before the break in the `increment` case is done at the return step.

It just keeps the "same" logic inside the loop, nothing else. A matter of taste, and also both solutions are possible.

gph/src/ripser.h
ulupo: I'm confused about this, and particularly the implications for taking the square and cube roots.  Should we just keep the types what they are and hope that `std::sqrt` and `std::cbrt` deal with integer types in the best way possible?

gph/src/ripser.h
ulupo: Ok!  I'll go with the first one for now just because it seems more human-friendly to me.

gph/src/ripser.h
MonkeyBreaker: That is a good question, when looking at the documentation of both function [sqrt](https://en.cppreference.com/w/cpp/numeric/math/sqrt) and [cbrt](https://en.cppreference.com/w/cpp/numeric/math/cbrt), they both have a special input for any integer types, and if you look at a note, it indicates the following:

> A set of overloads or a function template accepting an argument of any integral type. Equivalent to 2) (the argument is cast to double).

It seems that when you input an integral type, it is implicitly casted to `double`.

I am not a big fan of `long double` because it is architecture dependent on the results provided. If we let `long double` 2 different users with the same input could lead to different results, because they have a different architecture, it will be a quite hard behavior to debug.

See [link](https://stackoverflow.com/a/14221894) for a detailed explanation on the format differences.

gph/src/ripser.h
ulupo: Done.

gph/src/ripser.h
ulupo: Done!

gph/src/ripser.h
ulupo: Ok!  So it seems we are looking for a fast, exact implementation of the "integer square root" and "integer cube root".  I'm sure someone has done this!

ulupo: https://stackoverflow.com/a/31120562

MonkeyBreaker: Nice, but we will need to look and test carefully this kind of solution, but I like that :smile:  !
I can have a look in my book "Hacker's delight" if there is a suggestion for this kind of misc functions.

ulupo: I can't think of a better person than you to help me figure this out!

MonkeyBreaker: Thank you, will try to do my best with my current availability ...

ulupo: Boost! http://rosettacode.org/wiki/Isqrt_(integer_square_root)_of_X#C.2B.2B

ulupo: Other sources:
  - https://codereview.stackexchange.com/a/75345 (integer square root)
  - https://stackoverflow.com/a/4331882 (integer cube root)

gph/src/ripser.h
```diff
@@ -706,11 +731,18 @@ class ripser
                                         index_t n, OutputIterator out) const
     {
         --n;
-        for (index_t k = dim + 1; k > 0; --k) {
+        for (index_t k = dim + 1; k > 2; --k) {
```
MonkeyBreaker, line 734: I understand why you change the condition to `k > 2` but could you just a short comment.
To shortly explains it is because the last 2 remaining vertices are computed differently just after the loop ?

ulupo:
We are currently facing the following issues:
  1. we will experience overflows towards the limit of the `index_t` range because of the multiplications by `2` or `6`;
  2. we don't really know if the floating-point computation of the square root (followed by flooring or rounding) is guaranteed to lead to the correct result.

Problem 1 and the fact that floating-point operations are *so fast* can mean that it's not such a bad idea to stick to `std::sqrt` and use floating-point square roots.
Problem 2 can be addressed in at least one of two ways:
  - *test* the correctness by brute force on all the `index_t` range? (Portability? Robustness to 128-bit implementations?)
  - adopt a similar approach as in 2D: treat the floating-point computation as leading to a first guess, then confirm or reject the guess by a linear search. (Performance?)

ulupo:
> test the correctness by brute force on all the index_t range. (Portability? Robustness to 128-bit implementations?)

This is unfeasible, we would need 10^8--10^9 minutes of runtime :)

ulupo:
> I think I can demonstrate that it works with (and probably until) 2^27 vertices, but is off by 1 for 2^28 vertices and more (i.e. until approximately 2^32).

In view of this finding, it seems that in the case of 1D we should also try out a provably correct alternative which refines the guess as in the 1D case, and hope that the performance impact is small.

gph/src/ripser.h
ulupo: @MonkeyBreaker I have done a preliminary test and replacing `long double` with just `double` does seem to yield better performance.  So maybe another argument to remove `long double`.

MonkeyBreaker: Well, for me it was not related to performances, but to inconsistent behavior with the same input data when run on 2 different architectures.
I would take with a grain of salt any performance difference between `double` and `long double`

ulupo: Right. For 2-simplices we cannot ensure the numerics will work exactly even for relatively small numbers, unlike the case of 1-simplices. So my point is that if using `double` still implies we are only 1 or two steps away from the correct results, and if it can be faster on some architectures, then it should be preferred.

ulupo: > if using double still implies we are only 1 or two steps away from the correct results for a wide range of indices

And I think this is true at least for datasets with as many as one million points.

gph/src/ripser.h
```diff
@@ -691,9 +691,34 @@ class ripser
...
+                    if (!pred(guess)) {
+                        --guess;
+                        break;
+                    }
+                } else {
+                    --guess;
+                    if (pred(guess))
+                        break;
+                }
+            }
```
ulupo, line 718: We need to give more careful thought about edge cases: for example we mustn't increment if `guess` equals `n - 1` or else we go out of bounds, and we can't decrement if `guess` equals 0.  Maybe this is ensured by `pred` already?

gph/src/ripser.h
ulupo: I have added these comments now.

MonkeyBreaker: Perfect, I resolve the conversation then

ulupo:
Note: one can prove that floor((sqrt(8i+1)-1)/2) = (isqrt(8i+1)-1)/2 where isqrt(x)=floor(sqrt(x)), i.e. the integer square root.  So we could make the 1D optimization perfectly exact and not worry about floating point precision at all if:
  - we had a *fast* implementation of isqrt
  - limited the scope of the optimization to indices i such that 8i + 1 is in the `index_t` range, which is still lots (up to ~ 2^60 which would be good for input with over a billion vertices)

MonkeyBreaker:
Benchmarking time, all timing measured in single thread, no collapser and no generators.

| dataset | dim | thresh | time normal [s] | time ulupo [s] |
|---------|-----|--------|-----------------|----------------|
| sphere  | 2   | -      | 0.58            | 0.5            |
| dragon  | 1   | -      | 1.95            | 1.41           |
| random  | 7   | -      | 3.4             | 3.9            |
| fractal | 2   | -      | 4.55            | 2.9            |
| o3_1024 | 3   | 1.8    | 2               | 1.68           |
| o3_4096 | 3   | 1.4    | 37              | 30.5           |
| torus   | 2   | 0.15   | 45              | 29             |

Results are really quite impressive. The bigger the dataset the more it sames to provide speed-up.
But, `random` dataset is less performant. This dataset is the only one that does go to higher dimension than 3.
We are maybe paying the price of a linear search instead of a binary search in higher dimensions ?

DISCLAIMER, I did not test if the barcode produced are still good, this will be my task for the weekend.

ulupo:
@MonkeyBreaker wow, this is great! Out of curiosity, did you use the last version with `double` instead of `long double`?  I think random has very few points so the binary search would take very few iterations and thus becomes competitive with taking square roots or cube roots.

Another interesting test would be to remove the optimization for 2-simplices (so no linear search) and only leave the one for 1-simplices in place.

MonkeyBreaker:
@ulupo I used commit 7c1dfd1, the latest I think. So I think it uses `double`.
I can try to remove the 2-simplicies optimization, but not right now.

Makes sense for the `random` case, I will try with other small dataset to increase `maxdim` to verify if this could be the reason.

ulupo:
@MonkeyBreaker I think we should resurrect this PR but am proposing we restrict its objective to the implementation of only one kind of speedup, namely the use of floating-point square root for speeding up the computation in the case of 1-simplices only.  So I reverted some changes and brought the diff to a very simple state.

There is another family of speedups worth considering, based on integer operations alone.  On 1-simplices, they are not as good as this floating-point speedup (according to tests I ran), but they generalize to higher-dimensional simplices, and I have observed that they lead to speedups in some cases. I have created many variants of these other speedups, and have a personal favourite, but the gains are less impressive than the ones provided by this PR, while their implementation is considerably more complex.  So I propose that we study those in a separate PR if and after this one is merged.

I ran some benchmarks and report the values I obtained below:

| dataset | dim | thresh | time normal [s] | time ulupo [s] |
|---------|-----|--------|-----------------|----------------|
| sphere  | 2   | -      | 748 ms ± 22.3 ms            | 643 ms ± 22.2 ms           |
| dragon  | 1   | -      | 2.09 s ± 23.4 ms            | 1.84 s ± 32.4 ms           |
| random  | 7   | -      | 4.85 s ± 20.1 ms             | 4.86 s ± 53.5 ms            |
| fractal | 2   | -      | 6.28 s ± 29.1 ms            | 4.64 s ± 20.2 ms            |
| o3_1024 | 3   | 1.8    | 2.78 s ± 25.9 ms               | 2.59 s ± 19.7 ms           |
| o3_4096 | 3   | 1.4    | 50.4 s ± 214 ms             | 48.3 s ± 338 ms           |
| torus   | 2   | 0.15   | 52 s ± 287 ms              | 42.6 s ± 178 ms            |

About numerical issues: a similar use of the floating-point square root is made already by ripser, when computing the number of rows of the compressed distance matrix.  Hence, we already implicitly accepted the use of floating-point arithmetic, and trusted that this did not lead to errors due to numerics.  Above in this thread I collected more rigorous thoughts about why this should work anyway (up to a certain maximum number of edges which will be more than high enough for our users).  I'm happy to incorporate some of these thoughts as inline comments.

ulupo:
@MonkeyBreaker on the other hand, I am quite worried to observe that performance has degraded substantially since a previous version of the main branch.  My previous values on `main` had been as follows:

| dataset | dim | thresh | time normal [s] |
|---------|-----|--------|-----------------|----------------|
| sphere  | 2   | -      | 484 ms ± 1.81 ms            |
| dragon  | 1   | -      | 1.36 s ± 4.73 ms            |
| random  | 7   | -      | 3.21 s ± 20.2 ms             |
| fractal | 2   | -      | 4.19 s ± 4.84 ms            |
| o3_1024 | 3   | 1.8    | 1.81 s ± 4.03 ms               |
| o3_4096 | 3   | 1.4    | 32.9 s ± 179 ms             |
| torus   | 2   | 0.15   | 35 s ± 274 ms              |

MonkeyBreaker:
@ulupo, looking at the issue you mentioned and as we discussed together.

On the other side, the benchmark you requested on my machine:

| dataset   | dim   | thresh   | giotto-ph latest (23b016) [s] | giotto-ph specialized [s] |
| --------- | ----- | -------- | --------------------          | --------------------      |
| sphere    | 2     | -        | 0.54                          | 0.46                      |
| dragon    | 1     | -        | 1.53                          | 1.41                      |
| random    | 7     | -        | 3.26                          | 3.32                      |
| fractal   | 2     | -        | 3.96                          | 2.37                      |
| o3_1024   | 3     | 1.8      | 1.90                          | 1.75                      |
| o3_4096   | 3     | 1.4      | 42.1                          | 38.4                      |
| torus     | 2     | 0.15     | 46.7                          | 38.8                      |

I run each dataset 5 times and I report the minimum timing of the five run. As we discuss, repeating the same dataset right now has a regression in performances.

It is impressive for Fractal the results obtained with this small change. The only dataset it becomes slightly slower is with random.

MonkeyBreaker (approved):

MonkeyBreaker:
Hey @ulupo,

As discussed, you would like to integrate this version of the speed-up vertices enumeration.
IMO, all test are passing and the results on my machine look good.

The only point I am unsure, is that you introduce `sqrt` operation, and this may be costly on processors that do not have a specialized instruction for computing it. But, I don't think that any modern processor are not equipped with this kind of instruction.

If you would like to add anything, let me know, otherwise let's go with merging this !

Issue #39, opened 2021-11-01
maxdim=0


mglisse:
Hello,
I tried to run `gph.ripser_parallel(np.array([[1.,2],[3,4],[5,0]]), maxdim=0)` but it segfaults (while it works with maxdim=1).

MonkeyBreaker:
Hello,

Thank you for reporting the issue. After debugging what was going on, I found out a fix.
Currently a PR (#40) is in progress for patching this issue.

Sorry that you encountered a bug and thank you for reporting it with an example.

Best,
Julián

MonkeyBreaker:
The fix is now merged but currently not available in Pypi.
We will soon release version 0.2.0, but on the meantime if you want to work with `giotto-ph` with `maxdim=0` fix, you can follow the [developer installation](https://giotto-ai.github.io/giotto-ph/build/html/installation.html#developer-installation).

PR #40, opened 2021-11-02
Fix issue when computing PH with maxdim == 0

MonkeyBreaker:
This PR fixes issue #39 when computing persistent homology with the maximal dimension set to 0.

The issue was that in function `compute_dim_0`, we also prepare the columns to reduce for dimensions 1. But, when `maxdim=0`, we don't care about preparing those columns. Even worse was that `get_zero_apparent_cofacet(e, 1)` was assuming that we were always computing for `maxdim>0`.

This is now fixed by preparing the columns for dimension only if `maxdim > 0`.

### DIFF
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -321,8 +321,5 @@ def test_ph_maxdim_0():
     res_maxdim_0 = ripser(X, maxdim=0)['dgms'][0]
     res_maxdim_1 = ripser(X, maxdim=1)['dgms'][0]
 
-    # Verifies that the number of barcodes is the same
-    assert res_maxdim_0.shape[0] == res_maxdim_1.shape[0]
-
-    # Verifies if both computation have the same barcodes
+    # Verifies if both computations have the same barcodes
     assert_array_equal(res_maxdim_0, res_maxdim_1)
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -315,6 +315,15 @@ def test_gens_order_vertices_higher_dimension():
 
 def test_ph_maxdim_0():
     """Regression test for issue #39, an issue was found when only computing
-    up to dimension 0. The test also compares the results of dimension 0"""
+    up to dimension 0. The test also compares the results of dimension 0
+    when maxdim=1 and maxdim=0"""
     X = np.array([[1., 2], [3, 4], [5, 0]])
-    ripser(X, maxdim=0)['dgms'][0]
+    res_maxdim_0 = ripser(X, maxdim=0)['dgms'][0]
+    res_maxdim_1 = ripser(X, maxdim=1)['dgms'][0]
+
+    # Verifies that the number of barcodes is the same
+    assert res_maxdim_0.shape[0] == res_maxdim_1.shape[0]
+
+    # Verifies if both computation have the same barcodes
+    for barcode in res_maxdim_0:
+        assert any(np.equal(barcode, res_maxdim_1).all(1))
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -311,3 +311,10 @@ def test_gens_order_vertices_higher_dimension():
 
     assert len(gens_fin_dim2) == 1
     assert np.array_equal(gens_fin_dim2[0], np.array([1, 0, 5, 4]))
+
+
+def test_ph_maxdim_0():
+    """Regression test for issue #39, an issue was found when only computing
+    up to dimension 0. The test also compares the results of dimension 0"""
+    X = np.array([[1., 2], [3, 4], [5, 0]])
+    ripser(X, maxdim=0)['dgms'][0]
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -3,7 +3,7 @@ import pytest
 from hypothesis import given, settings
 from hypothesis.extra.numpy import arrays
 from hypothesis.strategies import floats, integers, composite
-from numpy.testing import assert_almost_equal
+from numpy.testing import assert_almost_equal, assert_array_equal
 from scipy.sparse import coo_matrix
 from scipy.spatial.distance import pdist, squareform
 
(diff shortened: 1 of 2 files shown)


gph/python/test/test_ripser.py
```diff
@@ -311,3 +311,10 @@ def test_gens_order_vertices_higher_dimension():
 
     assert len(gens_fin_dim2) == 1
     assert np.array_equal(gens_fin_dim2[0], np.array([1, 0, 5, 4]))
+
+
+def test_ph_maxdim_0():
+    """Regression test for issue #39, an issue was found when only computing
+    up to dimension 0. The test also compares the results of dimension 0"""
```
ulupo, line 318: I guess you originally meant to say "The test also compares the results of dimension 0" when `maxdim=1` and `maxdim=0`?

gph/src/ripser.h
```diff
@@ -1014,7 +1015,8 @@ class ripser
                         }
                     }
                 }
-            } else if (get_index(get_zero_apparent_cofacet(e, 1)) == -1)
+            } else if (prepare_higher_dims &&
```
ulupo, line 1018: Is this another case of branch prediction implying no overhead from doing this every time we see an edge rather than deciding at the beginning?

ulupo (changes requested):

gph/src/ripser.h
MonkeyBreaker: We help the branch predictor. I don't know for sure if this would results in a speed-up.
And also, in my opinion, it is much clear why we add this check.

gph/python/test/test_ripser.py
MonkeyBreaker: That was what I had previously ..........
Let me remove this part

gph/src/ripser.h
ulupo: Ok great!  It's only dimension 0 anyway, so not exactly where we have to optimize to death.

gph/python/test/test_ripser.py
ulupo: ... or add the extra check in the test, which would be good also!

MonkeyBreaker: Let me add the check then ;)

MonkeyBreaker: Done

gph/python/test/test_ripser.py
```diff
@@ -311,3 +311,19 @@ def test_gens_order_vertices_higher_dimension():
...
+    X = np.array([[1., 2], [3, 4], [5, 0]])
+    res_maxdim_0 = ripser(X, maxdim=0)['dgms'][0]
+    res_maxdim_1 = ripser(X, maxdim=1)['dgms'][0]
+
+    # Verifies that the number of barcodes is the same
+    assert res_maxdim_0.shape[0] == res_maxdim_1.shape[0]
+
+    # Verifies if both computation have the same barcodes
+    for barcode in res_maxdim_0:
+        assert any(np.equal(barcode, res_maxdim_1).all(1))
```
ulupo, line 329: Why not just
```
np.assert_array_equal(res_maxdim_0, res_maxdim_1)
```
? One line for the whole thing (you need to add `assert_array_equal` to the imports at the top of the file; it should be imported together with `assert_almost_equal` from `numpy.testing`)

ulupo (changes requested):

gph/python/test/test_ripser.py
MonkeyBreaker: I did not know ^^
Let me commit your suggestion

MonkeyBreaker: Done, thank you for let me know about this function !

gph/python/test/test_ripser.py
```diff
@@ -311,3 +311,18 @@ def test_gens_order_vertices_higher_dimension():
...
+    up to dimension 0. The test also compares the results of dimension 0
+    when maxdim=1 and maxdim=0"""
+    X = np.array([[1., 2], [3, 4], [5, 0]])
+    res_maxdim_0 = ripser(X, maxdim=0)['dgms'][0]
+    res_maxdim_1 = ripser(X, maxdim=1)['dgms'][0]
+
+    # Verifies that the number of barcodes is the same
+    assert res_maxdim_0.shape[0] == res_maxdim_1.shape[0]
+
+    # Verifies if both computation have the same barcodes
```
ulupo, line 327:
```suggestion
    # Verifies if both computations have the same barcodes
```

gph/python/test/test_ripser.py
```diff
@@ -311,3 +311,18 @@ def test_gens_order_vertices_higher_dimension():
...
+def test_ph_maxdim_0():
+    """Regression test for issue #39, an issue was found when only computing
+    up to dimension 0. The test also compares the results of dimension 0
+    when maxdim=1 and maxdim=0"""
+    X = np.array([[1., 2], [3, 4], [5, 0]])
+    res_maxdim_0 = ripser(X, maxdim=0)['dgms'][0]
+    res_maxdim_1 = ripser(X, maxdim=1)['dgms'][0]
+
+    # Verifies that the number of barcodes is the same
+    assert res_maxdim_0.shape[0] == res_maxdim_1.shape[0]
```
ulupo, line 325: Redundant check now, you can remove it :)

ulupo (changes requested):

gph/python/test/test_ripser.py
MonkeyBreaker: Done

gph/python/test/test_ripser.py
MonkeyBreaker: Updated

ulupo (approved):
LGTM!

MonkeyBreaker:
Thank you for all the review Umbe, if all test passes, I will merge the PR.

ulupo:
@MonkeyBreaker sure! I prefer squash and merge for a cleaner history (not sure if this is what you prefer too)

MonkeyBreaker:
> @MonkeyBreaker sure! I prefer squash and merge for a cleaner history (not sure if this is what you prefer too)

I will then squash and merge, I prefer we stay consistent :smiley:

PR #41, opened 2021-11-02
Remove check for zero diameters in 0-dimensional code

ulupo:
Attempts to fix #31, or at least one of the issues presented there.

- [ ] Add unit/regression test
- [ ] Decide whether there really is an issue with birth time as inline comments suggest

PR #43, opened 2021-11-08
[CPP] Remove zero diameter check for edges in 0-dim computation

### LINKED ISSUE #31: [BUG] Dead comment in compute_dim_0_pairs/incorrect removal of edges with zero filtration value
Am i right to think that this comment

https://github.com/giotto-ai/giotto-ph/blob/c7f5d46dbf61c68b5006c888dff6bda5e3a528a3/gph/src/ripser.h#L936-L937

is referring to an issue that is now resolved?

ulupo:
Fixes #31, adding a unit test for "equivariance" (the property I was checking for in the example in #31) which fails in `main` but not after the fix.

New version of #41.

### DIFF
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -1,12 +1,13 @@
 import numpy as np
 import pytest
-from gph import ripser_parallel as ripser
 from hypothesis import given, settings
 from hypothesis.extra.numpy import arrays
 from hypothesis.strategies import floats, integers, composite
 from numpy.testing import assert_almost_equal, assert_array_equal
 from scipy.sparse import coo_matrix
 from scipy.spatial.distance import pdist, squareform
+
+from gph import ripser_parallel as ripser
 
 
 @composite
@@ -352,6 +353,31 @@ def test_equivariance(dm):
         for i, offset in enumerate(offset_cand):
             if not np.any(dm - offset == -offset):
                 offsets[i] = offset_cand
+
+    dgms_orig = ripser(dm, **kwargs)["dgms"]
+
+    for offset in offsets:
+        dgms_offset = ripser(dm - offset, **kwargs)["dgms"]
+        for dim in range(maxdim + 1):
+            assert_array_equal(dgms_offset[dim], dgms_orig[dim] - offset)
+
+
+def test_equivariance_regression():
+    """Make sure that, if `hypothesis` did not pick up a distance matrix like
+    the one in issue #31, then we test it by hand anyway, to avoid regressions.
+    """
+    maxdim = 1
+    kwargs = {"metric": "precomputed", "maxdim": maxdim}
+    dm = np.array([[0, 1, 3, 5, 4],
+                   [1, 0, 6, 7, 2],
+                   [3, 6, 0, 8, 9],
+                   [5, 7, 8, 0, 10],
+                   [4, 2, 9, 10, 0]], dtype=np.float32)
+
+    dm_flat = squareform(dm)
+    offsets = [np.median(dm_flat),
+               np.min(dm_flat),
+               np.max(dm_flat)]
 
     dgms_orig = ripser(dm, **kwargs)["dgms"]
 
--- a/gph/python/test/test_ripser.py
+++ b/gph/python/test/test_ripser.py
@@ -320,3 +320,42 @@ def test_ph_maxdim_0():
 
     # Verifies that the two computations lead to the same barcodes
     assert_array_equal(res_maxdim_0, res_maxdim_1)
+
+
+@settings(deadline=500)
+@given(dm=get_dense_distance_matrices())
+def test_equivariance(dm):
+    """Test that if we shift all entries (diagonal or not) in the input matrix
+    by a constant then the barcodes are also shifted by that constant. Doubles
+    also as a regression test for issue #31."""
+    maxdim = 1
+    kwargs = {"metric": "precomputed", "maxdim": maxdim}
+
+    # - Median may or may not make some edges zero, and will make some edges
+    #   negative and all vertex weights negative
+    # - Min is guaranteed to make at least one edge exactly zero, and all
+    #   vertex weights negative
+    # - Max is guaranteed to make all edge and vertex weights non-positive (or
+    #   infinite)
+    dm_flat = squareform(dm, checks=False)  # Get strict upper diagonal
+    dm_finite_nonzero_flat_sorted = np.sort(
+        dm_flat[np.logical_and(dm_flat > 0, dm_flat < np.inf)]
+        )  # Filter out zero values and infinite values, and sort
+    # This test will fail if val - offset = -offset numerically for some entry
+    # val in dm, because this will make some 0-dimensional classes disappear
+    # (they are born dead)
+    offsets = [0, 0, 0]  # Fallback if no edge weights are finite and positive
+    if len(dm_finite_nonzero_flat_sorted):
+        offset_cand = [np.float32(np.median(dm_finite_nonzero_flat_sorted)),
+                       np.min(dm_finite_nonzero_flat_sorted),
+                       np.max(dm_finite_nonzero_flat_sorted)]
+        for i, offset in enumerate(offset_cand):
+            if not np.any(dm - offset == -offset):
+                offsets[i] = offset_cand
+
+    dgms_orig = ripser(dm, **kwargs)["dgms"]
+
+    for offset in offsets:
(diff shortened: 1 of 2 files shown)


gph/python/test/test_ripser.py
```diff
@@ -1,14 +1,13 @@
 import numpy as np
 import pytest
+from gph import ripser_parallel as ripser
```
MonkeyBreaker, line 3: Should we not separate the import of the package being tested by the general import of externals libraries ?

gph/src/ripser.h
```diff
@@ -995,26 +995,25 @@ class ripser
...
-                                {birth_vertex.second, std::max(v1, v2),
-                                 std::min(v1, v2)});
-                        }
+                /* Unlike Ripser/ripser, here we do not exclude edges with
+                 * zero "diameters". This is because we don't assume that
+                 * all vertex weights are zero, as Ripser/ripser does. */
+                /* Elder rule; youngest class (max birth time of u and v)
+                 * dies first */
+                birth = birth_vertex.first;
+                value_t death = get_diameter(e);
```
MonkeyBreaker, line 1005: To be consistent, can we update the code has follow:

L.989: from `value_t birth;` to `value_t birth, death;`

And line L.1005: from `value_t death = get_diameter(e);` to `death = get_diameter(e);`

To be more consistent with the coding.

gph/src/ripser.h
```diff
@@ -995,26 +995,25 @@ class ripser
...
+                    births_and_deaths_by_dim[0].push_back(birth);
+                    births_and_deaths_by_dim[0].push_back(death);
+
+                    if (return_flag_persistence_generators) {
+                        flag_persistence_generators.finite_0.push_back(
+                            {birth_vertex.second, std::max(v1, v2),
+                             std::min(v1, v2)});
                     }
                 }
+
```
MonkeyBreaker, line 1016: Remove this empty line ?

gph/src/ripser.h
```diff
@@ -967,8 +967,8 @@ class ripser
     void compute_dim_0_pairs(std::vector<diameter_index_t>& edges,
                              std::vector<diameter_index_t>& columns_to_reduce)
     {
-        // TODO: Get correct birth times if the edges are negative (required for
-        // lower star)
+        /* TODO: Ensure https://github.com/scikit-tda/ripser.py/issues/135 is no
```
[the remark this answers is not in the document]
ulupo: Because I regard it as owned by ripser.py.  In any case, the issues are cross-linked.

gph/python/test/test_ripser.py
```diff
@@ -22,11 +21,9 @@ def get_dense_distance_matrices(draw):
                                      exclude_min=True,
                                      width=32),
                      shape=(shapes, shapes), unique=False))
-    # mirror along the diagonal the values of the
-    # distance matrix
-    dm = np.triu(dm.astype(np.float32), k=0)
+    # Extract strict upper diagonal and make symmetric
+    dm = np.triu(dm.astype(np.float32), k=1)
```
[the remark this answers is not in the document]
ulupo: Yes, and we avoid overflow warnings.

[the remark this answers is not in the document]
MonkeyBreaker: Perfect !

gph/src/ripser.h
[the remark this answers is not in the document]
MonkeyBreaker: Sure, I just found it weird to do a back and forth between 2 different repositories.
In our own issue, we already have the cross-reference to `ripser.py`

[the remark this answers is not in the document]
ulupo: Another reason: Issue #31 really became about the zero edges, while the ripser.py issue is more properly about the dead comment.

gph/python/test/test_ripser.py
MonkeyBreaker: Well, it sort them alphabetically, the issue for me is that we cannot differentiate current library import, or "internals" imports from external dependencies import.
I think it is important to have a clear difference between both

gph/src/ripser.h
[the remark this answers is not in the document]
MonkeyBreaker: Right ! You convinced me, I see the difference, sorry to not have the difference by my self !

gph/src/ripser.h
ulupo: Done.

gph/src/ripser.h
ulupo: Done.

ulupo:
@MonkeyBreaker the test is good, but unlike what I confidently stated above it does not fail (or fails so rarely I couldn't make it fail) on current `main`.  Evidently `hypothesis` does not produce something like my example in the discussion in #31 (for which I had to pick a specific random seed).

What I am thinking is:
  - keep the current test as is -- we like `hypothesis`
  - add my specific example by hand at the end of the test
    - make sure once again that this addition does fail in `main`

gph/python/test/test_ripser.py
ulupo: Makes sense!  I'll revert.

MonkeyBreaker:
> keep the current test as is -- we like hypothesis

Well we **must** be confident that our test fails if we break something.
I am not a fan of having test that won't fail at each time that we run if we broke the corresponding feature tested.

> add my specific example by hand at the end of the test 

Well, instead of having at the end of the test, implement 2 test, one with hypothesis that produces multiple tests.
One with the hard-coded input data that we know fails if the fix is not implemented.
I the input matrix is small, I would hard-code the matrix directly, because if we depend on the behavior of `np.seed()` it is not IMO reliable enough.

> make sure once again that this addition does fail in main

Sure, I can also cross-validate

gph/python/test/test_ripser.py
MonkeyBreaker: You can resolve the conversation afterwards, thank you !

ulupo:
> Well we must be confident that our test fails if we break something.
> I am not a fan of having test that won't fail at each time that we run if we broke the corresponding feature tested.

Not sure I follow -- this would definitely happen after my edits.  But it is the same test -- for the same property.  Testing with hypothesis covers tricky things like overflows, etc, while the hard-coded example makes sure we don't regress.  But it is the same property (equivariance) that is being tested.

ulupo:
On the other hand, I'm happy to split the tests -- e.g. because otherwise the same specific `dm` will be tested multiple times, one per run made by `hypothesis`.  This alone is a good reason for me to split!

gph/python/test/test_ripser.py
ulupo: Done.

ulupo:
@MonkeyBreaker I have now added the separate explicit regression test as discussed.  I checked that it fails on `main` currently.

Let me know if you'd like further changes for this PR.

gph/python/test/test_ripser.py
```diff
@@ -321,5 +319,69 @@ def test_ph_maxdim_0():
...
+    """Make sure that, if `hypothesis` did not pick up a distance matrix like
+    the one in issue #31, then we test it by hand anyway, to avoid regressions.
+    """
+    maxdim = 1
+    kwargs = {"metric": "precomputed", "maxdim": maxdim}
+    dm = np.array([[0, 1, 3, 5, 4],
+                   [1, 0, 6, 7, 2],
+                   [3, 6, 0, 8, 9],
+                   [5, 7, 8, 0, 10],
+                   [4, 2, 9, 10, 0]], dtype=np.float32)
```
ulupo, line 375: This is an easier-to-understand version of the matrix as in #31 in the sense that instead of the float values I put the ranks of the values -- the smallest off-diagonal value gets a 1, the second smallest a 2, etc.  It makes it easier to check things with pen and paper.

MonkeyBreaker:
> Not sure I follow -- this would definitely happen after my edits. But it is the same test -- for the same property. Testing with hypothesis covers tricky things like overflows, etc, while the hard-coded example makes sure we don't regress. But it is the same property (equivariance) that is being tested.

I am good that we have 2 separate test for validating a feature. I was afraid that with only 1 test (without the case we are always sure it fails), we could not really with certitude if all test are good.

But now, with both of the test implemented, I am happy :)

gph/python/test/test_ripser.py
MonkeyBreaker: Perfect, let me just try this on my computer

MonkeyBreaker:
LGTM !
I could reproduce the issue on my computer with your test and after adding the fix, all test passes !

MonkeyBreaker (approved):

PR #44, opened 2021-11-08
Prepare v0.2.0

ulupo:
Bump version to 0.2.0 and add release notes

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,3 +1,37 @@
+Release 0.2.0
+=============
+
+Major release bringing new features, performance enhancements, and bug fixes.
+
+Major Features and Improvements
+-------------------------------
+
+- ``ripser_parallel`` can now return "flag persistence generators", i.e. vertices and edges creating/destroying bars in the barcode (`#29 <https://github.com/giotto-ai/giotto-ph/pull/29>`_). A new Jupyter notebook illustrates its usage.
+- The computation of the enclosing radius has been sped up (`#28 <https://github.com/giotto-ai/giotto-ph/pull/28>`_).
+- ``ripser_parallel`` can now be imported directly from ``gph`` (`#36 <https://github.com/giotto-ai/giotto-ph/pull/36>`_).
+- A garbage collection step which was found to negatively impact runtimes with little memory benefit has been removed (`#33 <https://github.com/giotto-ai/giotto-ph/pull/33>`_).
+
+Bug Fixes
+---------
+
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -24,7 +24,7 @@
 author = 'Julian Burella Perez'
 
 # The full version, including alpha/beta/rc tags
-release = '0.0.1'
```
ulupo, line 27: Why was this 0.0.1?

gph/_version.py
```diff
@@ -1,4 +1,4 @@
-"""``giotto-ph`` is a Python library implementing algorithms for the
+"""``giotto-ph`` is a C++/Python library implementing algorithms for the
```
ulupo, line 1: Agree?

MonkeyBreaker: Sure, it make totally sense !

docs/source/conf.py
MonkeyBreaker: Because it was never updated I think,
I think that previously we retrieved this value by importing `gph`, the issue with this is that you need to install `gph` to build the documentation. And, in my opinion, we should not do that.

Not for this release, but for a future release I will try to make this clever, and not hard-code the value at 2 different places.

ulupo: We have a similar issue in `giotto-tda`!

RELEASE.rst
```diff
@@ -1,3 +1,37 @@
...
+---------
+
+- Sparse input with (signed) 64-bit row and column indices is now correctly dealt with by the Python interface (`#34 <https://github.com/giotto-ai/giotto-ph/pull/34>`_).
+- A bug causing segfaults to occur when ``maxdim=0`` was passed to the C++ backend has been fixed (`#40 <https://github.com/giotto-ai/giotto-ph/pull/40>`_).
+- An algorithmic error in dealing with edges with zero weight in the 0-dimensional computation has been fixed (`#43 <https://github.com/giotto-ai/giotto-ph/pull/43>`_).
+
+Backwards-Incompatible Changes
+------------------------------
+
+None.
```
MonkeyBreaker, line 24: Wait, does it still work with `from gph.python import ripser_parallel` ?

MonkeyBreaker: It seems that yes, it works, I do not know why I though otherwise

MonkeyBreaker (approved):

MonkeyBreaker:
LGTM !
When merged I will launch Wheels and Documentation generation !
Congrats :rocket:

RELEASE.rst
ulupo: Indeed! I had to check too

Issue #45, opened 2021-11-08
[CI] Produce wheels for Python 3.10 and Apple Silicon


ulupo:
As per title!

MonkeyBreaker:
We could also update the Title to include support to `Python 3.10 ` ?

ulupo:
> We could also update the Title to include support to Python 3.10 ?

Yes, thanks!

PR #47, opened 2021-11-08
[CI] Add Python version 3.10, make sure pip and setuptools are upgraded during the CI

ulupo:
Partially addresses #45, and ensures `pip` and `setuptools` are updated.  This is important as e.g. the correct installation of our dependencies depends on having these packages up-to-date (we similarly upgrade in `giotto-tda` and `pyflagser`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptool
+          python -m pip install --upgrade pip setuptools
           # The --upgrade and --upgrade-strategy eager flags ensure that pip will always install the latest allowed version of all the dependencies.
           python -m pip install --upgrade --upgrade-strategy eager -r requirements.txt
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
             # TODO: When updating c++ backend, use new date (yy-mm-dd)
           key: ${{ runner.os }}-ccache-${{ matrix.python-version }}-21-11-08
 
-      - name: Install dependencies
+      - name: Upgrade pip and setuptools
         run: |
           python -m pip install --upgrade pip setuptools
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
         include:
           - os: ubuntu-latest
             path: ~/.cache/pip
@@ -62,11 +62,12 @@ jobs:
         continue-on-error: true
         with:
           path: ${{ matrix.path_ccache }}
-            # TODO: When updating c++ backend, use new date
-          key: ${{ runner.os }}-ccache-${{ matrix.python-version }}-21-02-12
+            # TODO: When updating c++ backend, use new date (yy-mm-dd)
+          key: ${{ runner.os }}-ccache-${{ matrix.python-version }}-21-11-08
 
       - name: Install dependencies
         run: |
+          python -m pip install --upgrade pip setuptool
           # The --upgrade and --upgrade-strategy eager flags ensure that pip will always install the latest allowed version of all the dependencies.
           python -m pip install --upgrade --upgrade-strategy eager -r requirements.txt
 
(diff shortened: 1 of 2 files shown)


ulupo:
@MonkeyBreaker if this works, we could maybe merge it and then make another PR from a branch here (not in one of our forks) to verify whether the wheels can be built already or we really need to upgrade the manylinux version in view of SciPy's news: https://github.com/scipy/scipy/releases/tag/v1.7.2

.github/workflows/ci.yml
```diff
@@ -62,11 +62,12 @@ jobs:
...
-            # TODO: When updating c++ backend, use new date
-          key: ${{ runner.os }}-ccache-${{ matrix.python-version }}-21-02-12
+            # TODO: When updating c++ backend, use new date (yy-mm-dd)
+          key: ${{ runner.os }}-ccache-${{ matrix.python-version }}-21-11-08
 
       - name: Install dependencies
         run: |
+          python -m pip install --upgrade pip setuptools
           # The --upgrade and --upgrade-strategy eager flags ensure that pip will always install the latest allowed version of all the dependencies.
           python -m pip install --upgrade --upgrade-strategy eager -r requirements.txt
```
ulupo, line 72: I'm not exactly complaining about this line but since `python -m pip install -e .` will install all the latest dependencies, why do we add this line? `numpy`, `scipy` and `scikit-learn` (which are our dependencies) are not present in the base environment already, so installing them for the first time will also automatically fetch the latest possible version.

ulupo:
> @MonkeyBreaker sorry, I guess we need to mirror these changes into a new branch in giotto-ai to test whether the wheel building step would work with Python 3.10!

Or we can merge this as is (since the CI passes) and worry about whether we can build wheels for Python 3.10 after the merge?  After all, it's OK for us to test for Python 3.10 already now even if we don't yet produce wheels for it.

MonkeyBreaker:
I just started the building of the wheels on your branch https://github.com/ulupo/giotto-ph/actions/runs/1436504308
When touching the CI, I think it would be better to create a branch directly into origin, that way we can run the Action directly from here.

If the builds build fine, then yes, we can merge, otherwise I will debug what is going on :)

MonkeyBreaker:
The wheels for Python 3.6 to Python 3.10 can be now generated :tada: !
@ulupo do you want to address the comment about the dependencies now or should we do it in a later PR ?

.github/workflows/wheels.yml
```diff
@@ -27,6 +27,16 @@ jobs:
           CIBW_TEST_COMMAND: python -m pytest {package}/gph/python/test
           CIBW_TEST_REQUIRES: pytest hypothesis
 
+      - name: Build wheels for Python 3.10
```
ulupo, line 30:
```suggestion
      # SciPy do not ship manylinux2010 wheels for Python 3.10, so we too build ours for manylinux2014
      - name: Build wheels for Python 3.10
```

ulupo (commented):
If you're happy with this comment I am happy to merge this!

MonkeyBreaker (approved):

ulupo:
>  Are you sure that pip install -e . does pull latest version of the packages ?

@MonkeyBreaker see https://pip.pypa.io/en/stable/cli/pip_install/ (Satisfying Requirements).  So yes except for pre-releases sometimes.

MonkeyBreaker:
> @MonkeyBreaker see https://pip.pypa.io/en/stable/cli/pip_install/ (Satisfying Requirements). So yes except for pre-releases sometimes.

Alright, perfect, we could remove it now then, what do you think ?

ulupo:
> Alright, perfect, we could remove it now then, what do you think ?

I think yes!

.github/workflows/ci.yml
ulupo: Looks like we're going for it.  Maybe we can remove this whole step and move the upgrade pip setuptools line to the top of the next step also.

MonkeyBreaker:
Done, when the pipelines ends, then let's Merge !

.github/workflows/ci.yml
```diff
@@ -68,8 +68,6 @@ jobs:
       - name: Install dependencies
```
ulupo, line 68: What do you prefer? Changing the name of this step to "Upgrade pip and setuptools" or moving it to line 79?

MonkeyBreaker: We cannot move it, because the `pip install -e .` part is done for Linux and Mac together and for Windows specifically.
But changing the name is fine for me.

ulupo: Ok! I'm changing it.

MonkeyBreaker (approved):

MonkeyBreaker:
LGTM !

PR #48, opened 2021-11-08
[CI] Fix wheels after issues with pip version

MonkeyBreaker:
This PR fix the latest issues we encountered when trying to build the wheels.

The issue that was happening was due to on Python3.8, with the version of pip, it tried to compiles from sources scipy library (not expected behavior).
After debugging, we found out that updating the version of `cibuildwheels` this is now resolved.

## Work done

* [x] : Update `cibuildwheels` to v2.2.2
* [x] : Use `cibuildwheels` Github action now that it is available
* [x] : Disable `musllinux` wheels generation because they currently fail

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -18,8 +18,10 @@ jobs:
       - uses: actions/setup-python@v2
         name: Install Python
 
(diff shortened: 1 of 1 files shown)


ulupo (approved):
LGTM!

PR #49, opened 2021-11-08
Add information about Python 3.10 being supported

ulupo:
PR created pre-emptively under the assumptions that:
1. #47 gets merged
2. The wheel building is successful for Python 3.10.


ulupo:
@MonkeyBreaker what do you think about this plan?

MonkeyBreaker:
@ulupo Seems good, but I do not understand why do a separate PR. In my opinion all this could fit in the PR about adding support to `Python 3.10`.
But it is definitely a good thing that for the official 0.2.0 version, we add support to `Python 3.10`

ulupo:
It's to separate a PR about CI from an easy PR about release notes and text, since the situation is so delicate.

MonkeyBreaker (approved):

PR #50, opened 2021-11-14
Small codebase updates

MonkeyBreaker:
This PR addresses multiples small issues here and there in the code base that where on my to-do list:

* [x] : Add templates to apparent pair related code to remove code logic duplication
* [x] : Add verification of coefficient parameter. Checks that the parameter is lower than the maximal value supported and if the value is a prime number
* [x] : Update naming of variables inside of `binomial_coeff_table` as requested by @ulupo
* [x] : Update barcodes returned from the bindings to directly be of Numpy type, because we want to match `float` 32-bit precision used instead of Python default one `float` 64-bit precision. The return value is therefore composed of Numpy arrays.
* [x] : Fixed an issue with recent version of GCC compiler with a header dependency that was removed from `algorithm` header.
* [x] : Remove factor parameter who was unused
* [x] : Remove `num_edges` return value who is unused and if needed, can be easily computed in Python.

### DIFF
--- a/gph/bindings/ripser_bindings.cpp
+++ b/gph/bindings/ripser_bindings.cpp
@@ -14,6 +14,30 @@
 
 namespace py = pybind11;
 
+/* This function allows a conversion from a vector of vector into a
+ * vector of numpy arrays. This would be translated in Python into a
+ * List[np.array, ...]
+ */
+std::vector<py::array_t<value_t>>
+to_numpy_barcodes(std::vector<barcodes_t> barcodes)
+{
+    auto mat = std::vector<py::array_t<value_t>>(barcodes.size());
+
+    size_t i = 0;
+    for (auto& barcode : barcodes) {
+        auto arr = py::array_t<value_t>(
+            py::buffer_info(barcode.data(),
+                            sizeof(value_t),  // itemsize
+                            py::format_descriptor<value_t>::format(),
+                            barcode.size()  // shape
+                            ));
+
+        mat[i++] = arr;
+    }
+
+    return mat;
+}
+
 #if defined USE_COEFFICIENTS
 PYBIND11_MODULE(gph_ripser_coeff, m)
 {
@@ -35,8 +59,10 @@ PYBIND11_MODULE(gph_ripser, m)
     // ImportError: generic_type: type "ripserResults" is already registered!
     // When same python module imports gtda_ripser and gtda_ripser_coeff
     py::class_<ripserResults>(m, "ripserResults", py::module_local())
-        .def_readwrite("births_and_deaths_by_dim",
-                       &ripserResults::births_and_deaths_by_dim)
+        .def("births_and_deaths_by_dim",
+             [&](ripserResults& res) {
+                 return to_numpy_barcodes(res.births_and_deaths_by_dim);
+             })
         .def_readwrite("flag_persistence_generators_by_dim",
                        &ripserResults::flag_persistence_generators)
         .def_readwrite("num_edges", &ripserResults::num_edges);
--- a/gph/bindings/ripser_bindings.cpp
+++ b/gph/bindings/ripser_bindings.cpp
@@ -78,8 +78,6 @@ PYBIND11_MODULE(gph_ripser, m)
                 compressed_lower_distance_matrix(
                     compressed_upper_distance_matrix(std::move(distances)));
 
-            // TODO: This seems like a dummy parameter at the moment
-            float ratio = 1.0;
             int num_edges = 0;
 
             for (auto d : dist.distances) {
@@ -88,10 +86,9 @@ PYBIND11_MODULE(gph_ripser, m)
             }
 
             ripserResults res;
-            ripser<compressed_lower_distance_matrix> r(std::move(dist), dim_max,
-                                                       threshold, ratio,
-                                                       modulus, num_threads,
-                                                       return_generators);
+            ripser<compressed_lower_distance_matrix> r(
+                std::move(dist), dim_max, threshold, modulus,
+                num_threads, return_generators);
             r.compute_barcodes();
             r.copy_results(res);
             res.num_edges = num_edges;
@@ -109,12 +106,10 @@ PYBIND11_MODULE(gph_ripser, m)
             auto J_ = static_cast<index_t*>(J.request().ptr);
             auto V_ = static_cast<value_t*>(V.request().ptr);
 
-            // TODO: This seems like a dummy parameter at the moment
-            const float ratio = 1.0;
             // Setup distance matrix and figure out threshold
             ripser<sparse_distance_matrix> r(
                 sparse_distance_matrix(I_, J_, V_, NEdges, N, threshold),
-                dim_max, threshold, ratio, modulus, num_threads,
+                dim_max, threshold, modulus, num_threads,
                 return_generators);
             r.compute_barcodes();
             // Report the number of edges that were added
--- a/gph/bindings/ripser_bindings.cpp
+++ b/gph/bindings/ripser_bindings.cpp
@@ -64,8 +64,7 @@ PYBIND11_MODULE(gph_ripser, m)
                  return to_numpy_barcodes(res.births_and_deaths_by_dim);
              })
         .def_readwrite("flag_persistence_generators_by_dim",
-                       &ripserResults::flag_persistence_generators)
-        .def_readwrite("num_edges", &ripserResults::num_edges);
+                       &ripserResults::flag_persistence_generators);
 
     m.def(
         "rips_dm",
@@ -78,20 +77,12 @@ PYBIND11_MODULE(gph_ripser, m)
                 compressed_lower_distance_matrix(
                     compressed_upper_distance_matrix(std::move(distances)));
 
-            int num_edges = 0;
-
-            for (auto d : dist.distances) {
-                if (d <= threshold)
-                    ++num_edges;
-            }
-
             ripserResults res;
             ripser<compressed_lower_distance_matrix> r(
                 std::move(dist), dim_max, threshold, modulus,
                 num_threads, return_generators);
             r.compute_barcodes();
             r.copy_results(res);
-            res.num_edges = num_edges;
             return res;
         },
         "D"_a, "N"_a, "modulus"_a, "dim_max"_a, "threshold"_a, "num_threads"_a,
@@ -112,16 +103,9 @@ PYBIND11_MODULE(gph_ripser, m)
                 dim_max, threshold, modulus, num_threads,
                 return_generators);
             r.compute_barcodes();
-            // Report the number of edges that were added
-            int num_edges = 0;
-            for (int idx = 0; idx < NEdges; idx++) {
-                if (I_[idx] < J_[idx] && V_[idx] <= threshold) {
-                    num_edges++;
-                }
-            }
+
             ripserResults res;
             r.copy_results(res);
-            res.num_edges = num_edges;
             return res;
         },
         "I"_a, "J"_a, "V"_a, "NEdges"_a, "N"_a, "modulus"_a, "dim_max"_a,
--- a/gph/bindings/ripser_bindings.cpp
+++ b/gph/bindings/ripser_bindings.cpp
@@ -132,4 +132,7 @@ PYBIND11_MODULE(gph_ripser, m)
         "I"_a, "J"_a, "V"_a, "NEdges"_a, "N"_a, "modulus"_a, "dim_max"_a,
         "threshold"_a, "num_threads"_a, "return_generators"_a,
         "ripser sparse distance matrix");
+
+    m.def("get_max_coefficient_field_supported",
+          []() { return (uintptr_t(1) << num_coefficient_bits) - 1; });
 }
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -520,6 +520,8 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                                    return_generators)
 
     # Unwrap persistence diagrams
+    # He must do a call now, because call a method of object ripserResults
+    # Instead of calling an attribute of the object
     dgms = res.births_and_deaths_by_dim()
     for dim in range(len(dgms)):
         N = int(len(dgms[dim]) / 2)
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -251,6 +251,21 @@ def _ideal_thresh(dm, thresh):
     return min([enclosing_radius, thresh])
 
 
+def _supported_coeff_field(coeff):
+    max_coeff = gph_ripser.get_max_coefficient_field_supported()
+
+    # https://stackoverflow.com/questions/2068372/fastest-way-to-list-all-primes-below-n-in-python/3035188#3035188
+    def rwh_primes(n):
+        """ Returns  a list of primes < n """
+        sieve = [True] * n
+        for i in range(3, int(n**0.5)+1, 2):
+            if sieve[i]:
+                sieve[i*i::2*i] = [False]*int(((n-i*i-1)/(2*i)+1))
+        return [2] + [i for i in range(3, n, 2) if sieve[i]]
+
+    return coeff < max_coeff and coeff in rwh_primes(max_coeff)
+
+
 def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                     metric_params={}, weights=None, weight_params=None,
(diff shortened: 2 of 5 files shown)


MonkeyBreaker:
@ulupo if you think of anything else that should be added to this PR.
 Please let me know, I am not sure additional things can/should be added on this PR, it addresses small things I got in my to-do list.

gph/python/test/test_ripser.py
```diff
@@ -235,7 +235,7 @@ def test_gens_edge_in_dm_and_sorted():
 def test_gens_with_collapser():
     """This test ensures that you cannot use collapser and
     retrieve representative simplices. This is a temporary behavior."""
-    X = squareform(pdist(np.random.random((100, 3))))
+    X = squareform(pdist(np.random.random((10, 3))))
```
MonkeyBreaker, line 238: I did this small update because we don't care about the input data, I don't want that the test are long just because we loose time in generating the input data when we are not interested in the results of the computation themself.

gph/src/ripser.h
```diff
@@ -92,7 +93,7 @@ typedef uint16_t coefficient_t;
 
 using barcodes_t = std::vector<value_t>;
 
-static const size_t num_coefficient_bits = 8;
+const size_t num_coefficient_bits = 8;
```
ulupo, line 96: In general, when do we want const and when do we want static const?

ulupo (commented):
Great effort, thank you very much!  A quick comment before a full review: I was under the impression that Ripser/ripser implemented its own is_prime function in C++. What is the reason for not incorporating it here?

gph/src/ripser.h
MonkeyBreaker: Good question.
When using static, the variable can only be accessed within the file itself, or compilation unit.
It could be similar to a private attribute for a Class, the static variable is can only be directly accessed within the file where it is declared.

MonkeyBreaker:
>I was under the impression that Ripser/ripser implemented its own is_prime function in C++. What is the reason for not incorporating it here

Well, the function `is_prime` was removed from the `ripser` source file.
In my opinion, a `is_prime`function should be more in a utility file than inside of the `ripser` PH computation file.
The only reason I would think to have the C++ `is_prime` function back, is if this becomes a bottleneck computation.
@ulupo, what do you think ?

gph/src/ripser.h
```diff
@@ -122,23 +123,23 @@ class binomial_coeff_table
...
         }
     }
 
-    index_t operator()(index_t k, index_t n) const
+    index_t operator()(index_t n, index_t k) const
     {
-        assert(n < B.size() && k < B[n].size() && n >= k - 1);
-        return B[n][k];
+        assert(k < B.size() && k < B[k].size() && n >= k - 1);
+        return B[k][n];
```
ulupo, line 142: Could we please also rename the `i`s to `j`s and vice-versa? That way, `j` is a column index and `i` is a row index, which I think is more standard than the other way around.

ulupo (changes requested):

gph/src/ripser.h
MonkeyBreaker: Well, by convention in programming, we start with letter `i` in nested for loops and then we go by `j`, etc.
I think if the name of the indices are important, then we should rename then `idx_row` and `idx_col`, because otherwise in my opinion it is confusing.

ulupo: I do think the code is better understood by others if we swap the names, even if this breaks the nested for-loop convention.  Which `idx_row` and `idx_col` are you referring to?

MonkeyBreaker: I understand the change, I still think it is not a good idea to swap `i` and `j` as names for the nested loops.
What I was suggesting is that if those name are relevant, instead of just being loops indices, we can replace as follow:
* `j` => `idx_col`
* `i` => `idx_row`

Otherwise, I am not favorable for the requested change, I think this will be confusing for anyone looking at the code to have loop using `i` as index name be inside of loop using `j` as name.

ulupo: Ah I see!  Well I guess we have different perspectives on this.  Those new names are a bit verbose, which I think impacts readability badly in another way.  And I can't think of better ones (I guess `r` and `c` are off the table?).  I'm happy to give this one up!

MonkeyBreaker: It happens :)

On contrary, `r` and `c` are good candidates. They are not usual names for loops indices, I am definitely fine with this change. If I am right, I should proceed with the following change:
* `j` => `c`
* `i` => `r`

Right ? If yes, I will add just a short comment to indicate that `c` is for column index and `r` is for row index.

ulupo: Thinking about it a bit more, I think the next best option is to keep the names as they are but add a short comment saying that `B[j][i] = binom(i, j)`.

ulupo: (This comment can be added to the pre-existing comment above, on transposition.)

MonkeyBreaker: I will add it then !
Thank you for the suggestion @ulupo and I sorry to be a bit stubborn

gph/bindings/ripser_bindings.cpp
```diff
@@ -106,4 +132,7 @@ PYBIND11_MODULE(gph_ripser, m)
         "I"_a, "J"_a, "V"_a, "NEdges"_a, "N"_a, "modulus"_a, "dim_max"_a,
         "threshold"_a, "num_threads"_a, "return_generators"_a,
         "ripser sparse distance matrix");
+
+    m.def("get_max_coefficient_field_supported",
+          []() { return (uintptr_t(1) << num_coefficient_bits) - 1; });
```
ulupo, line 137: What number is this on Unix and Windows?

gph/python/test/test_ripser.py
```diff
@@ -235,7 +235,7 @@ def test_gens_edge_in_dm_and_sorted():
 def test_gens_with_collapser():
     """This test ensures that you cannot use collapser and
     retrieve representative simplices. This is a temporary behavior."""
```
ulupo, line 237:
```suggestion
    retrieve generators. This is a temporary behavior."""
```

gph/python/test/test_ripser.py
ulupo: Good catch, thanks!

gph/bindings/ripser_bindings.cpp
MonkeyBreaker: Which number do you mean ?
Currently because `num_coefficient_bits` is set to 8, it should return 255 on all platforms.

ulupo: Ok! In that case, why not just have a list with all primes up to that value in another file, and import it?   Will save the user computational time and it's not hard to maintain as I doubt we will ever need to offer support for larger primes.

ulupo: The list/set would be

2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131, 137, 139, 149, 151, 157, 163, 167, 173, 179, 181, 191, 193, 197, 199, 211, 223, 227, 229, 233, 239, 241, 251.

MonkeyBreaker: The computational time to compute the prime number up to N it not that important (if of course N is not too big).
We could hard-code this, but in my opinion, we should let it flexible.

The reason is that if other user clone the library and they need a bigger number than 251, they will need to update the list. But with current solution, they only need to change the number of bits, and the rest will be done automatically.

gph/src/ripser.h
```diff
@@ -122,23 +123,23 @@ class binomial_coeff_table
     /* Transposed binomial table
      * It's transposed because access where done over the rows and not the
      * columns */
```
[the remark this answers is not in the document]
ulupo: Can I commit this?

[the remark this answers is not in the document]
MonkeyBreaker: Sure, I missed it

gph/bindings/ripser_bindings.cpp
ulupo: I profiled this function and it takes about 16 µs to run on my machine.  Computing PH of a point cloud of 30 points takes 1.4 ms, so listing the primes would be about 1% of the runtime, and for 20 points the runtime of PH is around 500 µs (so primes is about 3% of the runtime).

I'm not saying it's a lot, but I think we can do better by stopping early in the computation: e.g.

```
import math

def is_prime(x, N):
    """Test whether x < N is prime."""
    if not x % 2:
        return False

    sieve = [True] * N
    for i in range(3, min(x, int(math.sqrt(N))) + 1, 2):
        if sieve[i]:
            sieve[i * i::2 * i] = [False] * int(((N - i * i - 1) / (2 * i) + 1))

    return sieve[x]
```
will stop almost immediately if the user inputs `3` (which would be the most common use case).

MonkeyBreaker: It makes sense to change `rwh_primes` to return a boolean to say if the coefficient is prime.
I did not use the method thinking about performances, but if we can avoid spending time unnecessary, why not.

Do you mind if I push your suggestion ?

ulupo: > Do you mind if I push your suggestion ?

Go for it! Beware that it gives `False` when the user passes `2` (but this is treated separately I think?).

gph/python/ripser_interface.py
```diff
@@ -520,7 +540,9 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                                    return_generators)
 
     # Unwrap persistence diagrams
-    dgms = res.births_and_deaths_by_dim
+    # He must do a call now, because call a method of object ripserResults
+    # Instead of calling an attribute of the object
```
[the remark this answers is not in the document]
MonkeyBreaker: I don't know, because for generators, we don't need to call a function.

gph/src/ripser.h
MonkeyBreaker: Comment added, if fine for you, I will let you resolve the conversation

gph/python/ripser_interface.py
```diff
@@ -251,6 +252,20 @@ def _ideal_thresh(dm, thresh):
     return min([enclosing_radius, thresh])
 
 
+def _is_prime_and_larger_than_2(x, N):
+    """Test whether 2 < x <= N is prime. Returns False when x is 2."""
+    if not x % 2 or x > N:
+        return False
+
+    sieve = [True] * N
```
ulupo, line 260:
```suggestion
    # https://stackoverflow.com/questions/2068372/fastest-way-to-list-all-primes-below-n-in-python/3035188#3035188
    sieve = [True] * N
```

gph/python/ripser_interface.py
```diff
@@ -444,6 +459,13 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             "`collapse_edges` and `return_generators`cannot both be True."
         )
 
+    if coeff != 2 and \
+            not _is_prime_and_larger_than_2(
+                coeff, gph_ripser.get_max_coefficient_field_supported()):
```
ulupo, line 464: Maybe store `gph_ripser.get_max_coefficient_field_supported()` into a variable so we don't call it twice?

gph/python/ripser_interface.py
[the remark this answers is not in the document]
ulupo: I don't fully understand.  My point is that once this is merged it does not matter what the code was like before, and it might just end up being confusing to keep this comment in the very long term (esp. for other developers).  Unless there is something confusing about the current code for *new* developers?

gph/src/ripser.h
ulupo: Great, saw it, thanks!

gph/python/ripser_interface.py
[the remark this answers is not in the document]
MonkeyBreaker: What my comment is trying to address is that now when you want the `barcodes`. You call method `births_and_deaths_by_dim`. But when you need the generator, here, you call an attribute directy `flag_persistence_generators_by_dim`.

But maybe the issue is the text in the comment itself, maybe something like:

```suggestion
    # Barcodes must match the inner type of C++ core filtration value.
    # We call a method from the bindings that returns the barcodes as
    # numpy arrays with np.float32 type
```

If you think that it is not necessary, we can also remove the comment. I let it more for future developers if they ask themself why a method is call instead of calling the attribute directly.

gph/python/ripser_interface.py
MonkeyBreaker: The computation time is almost null.
But it will definitely make the line smaller, let me do the change :+1:

MonkeyBreaker: Done, I let you resolve the conversation if it is good for you

ulupo: Thanks!

gph/python/ripser_interface.py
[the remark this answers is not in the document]
ulupo: Aaah, I see!  The new comment is great, would you be happy to commit that?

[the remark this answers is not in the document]
MonkeyBreaker: Sure !

gph/python/ripser_interface.py
```diff
@@ -251,6 +252,21 @@ def _ideal_thresh(dm, thresh):
...
+    """Test whether 2 < x <= N is prime. Returns False when x is 2."""
+    if not x % 2 or x > N:
+        return False
+
+    # https://stackoverflow.com/questions/2068372/fastest-way-to-list-all-primes-below-n-in-python/3035188#3035188
+    sieve = [True] * N
+    for i in range(3, min(x, int(math.sqrt(N))) + 1, 2):
+        if sieve[i]:
+            sieve[i * i::2 * i] = \
+                [False] * int(((N - i * i - 1) / (2 * i) + 1))
```
ulupo, line 265:
```suggestion
    sieve = [True] * (x + 1)
    for i in range(3, int(math.sqrt(x)) + 1, 2):
        if sieve[i]:
            sieve[i * i::2 * i] = \
                [False] * ((x - i * i) // (2 * i) + 1)
```

This version actually works when `x=N`, and makes the code faster/cleaner.

gph/src/ripser.h
```diff
@@ -1298,12 +1290,12 @@ class ripser
 
         // extra vector is a work-around inability to store floats in the
         // hash_map
-        std::atomic<size_t> last_diameter_index{0}, idx_essential{0};
-        entry_hash_map deaths;
-        deaths.reserve(columns_to_reduce.size());
+        std::atomic<size_t> idx_persistence{0}, idx_essential{0};
```
ulupo, line 1293: `last_diameter_index` makes me think that this stores the edge indices of the largest edge within a simplex.  Is this correct?

MonkeyBreaker: `last_diameter_index` was really a bad name.
During the computation, this was storing the number of finite barcodes in the current dimension.
We were incrementing this value each time a new finite barcode was found, and we use the pre incremented value as an index to store in another vector the death value.

This is currently a bit messy, but it is due to a current limitation of the hash map to store floating values.
I hope that at some point we will do a new concurrent lock-free hash_map that supports float values.

ulupo: Thanks for the explanation!  `What would you think about `idx_finite_bar` for this variable?  Because `persistence` for me is a death - birth value.

MonkeyBreaker: I am all good for suggestions, it is a draft to try to improve variable names.
Any feedback from experts (like you ;) ) in the field is more than welcome !

MonkeyBreaker: Also, to align with another variable, what do you think about changing `new_bar_pers_idx` to `new_finite_idx` ?

ulupo: > Also, to align with another variable, what do you think about changing new_bar_pers_idx to new_finite_idx ?

I agree about the need to align, but then I would suggest new_idx_finite_bar if we go for idx_finite_bar.

MonkeyBreaker: Good for me, I will push an update and if everything else is good for you, I think we can merge !

MonkeyBreaker: If good for you, I let you resolve the conversation

ulupo: Thanks!

gph/src/ripser.h
ulupo: Thanks!

ulupo (approved):
Great work, thanks!

PR #51, opened 2021-11-15
[CPP] Avoid unnecessary binary search when k = 1 (0-simplices)

ulupo:
Since `k = 1` corresponds to 0-simplices, and by Lemma 4.1. in the [Ripser paper](https://arxiv.org/pdf/1908.02518.pdf), one has that `get_max_vertex(idx, 1, n) = max{i | (i choose 1} ≤ idx} = idx` and we do not need to perform any binary search.

This has a very non-trivial impact on performance: every  time `get_simplex_vertices` is called (for any homology dimension), we end up calling `get_max_vertex(idx, 1, n)` and this binary search takes just as long as any of the others.

The fix here was already implemented in #38 but only by accident.  I did not know how significant it was.  In fact I wonder if it might explain *most* of the performance gains introduced by that PR (?). For this reason I think this fix deserves to be merged before any other PR that attempts to go even further.

### DIFF
--- a/gph/src/ripser.h
+++ b/gph/src/ripser.h
@@ -706,11 +706,13 @@ public:
                                         index_t n, OutputIterator out) const
     {
         --n;
-        for (index_t k = dim + 1; k > 0; --k) {
+        for (index_t k = dim + 1; k > 1; --k) {
             n = get_max_vertex(idx, k, n);
             *out++ = n;
             idx -= binomial_coeff(n, k);
         }
+        // k = 1 is 0-simplices (vertices), and get_max_vertex(idx, 1, n) = idx
+        *out = idx;
         return out;
     }
 


MonkeyBreaker:
Seems good to me !
If you don't mind, let me do a benchmark tonight, so we have an approximation of the speed-up provided by this change !
@ulupo Sounds good for you ?

gph/src/ripser.h
```diff
@@ -706,11 +706,13 @@ class ripser
...
     {
         --n;
-        for (index_t k = dim + 1; k > 0; --k) {
+        for (index_t k = dim + 1; k > 1; --k) {
             n = get_max_vertex(idx, k, n);
             *out++ = n;
             idx -= binomial_coeff(n, k);
         }
+        // k = 1 is 0-simplices (vertices), and get_max_vertex(idx, 1, n) = idx
+        *out = idx;
```
MonkeyBreaker, line 715: This has also a non-trivial impact. Compared to original implementation, when adding the last element, you still move the iterator by one.
Here we don't do it anymore, so I also think this add a certain amount of speed-up.

ulupo: I'd be curious to see the performance difference without `++` and with it!

MonkeyBreaker:
Below you will find a table with latest benchmarks. I think that we could say that in average we are around 8-9 percent faster. Which is really great for this kind of small change, good job @ulupo !
I think that a bit of templating could maybe provide a bit more of speed-up by removing the for loop when dim is 0.
If you don't mind waiting a bit on this, I could give it a try by end of November, @ulupo what do you think ?

| dataset | dim | thresh | normal timing [s] | new timing [s] | speed-up [%] |
|---------|-----|--------|-------------------|----------------|--------------|
| sphere  | 2   | -      | 0.58              | 0.53           | 0.08         |
| dragon  | 1   | -      | 1.95              | 1.59           | 0.18         |
| random  | 7   | -      | 3.4               | 3.28           | 0.03         |
| fractal | 2   | -      | 4.55              | 3.95           | 0.13         |
| o3_1024 | 3   | 1.8    | 2                 | 1.92           | 0.04         |
| o3_4096 | 3   | 1.4    | 37                | 34.7           | 0.06         |
| torus   | 2   | 0.15   | 45                | 40.7           | 0.09         |

ulupo:
Thanks for the benchmarks @MonkeyBreaker !! I'm glad there were no performance regression (as expected) and it's quite a lot for such a simple thing.

I see your point about templating, and I can certainly wait. On the other hand, I would have a slight preference towards merging this as is, so we can make more informed decisions moving forward in #38, and we can have a cleaner discussion there when making new benchmarks. What do you think?

MonkeyBreaker:
> I see your point about templating, and I can certainly wait. On the other hand, I would have a slight preference towards merging this as is, so we can make more informed decisions moving forward in #38, and we can have a cleaner discussion there when making new benchmarks. What do you think?

Well, I would wait that the CI pipeline works again with Python 3.10. If I can have something before that, lets include it in the this PR, otherwise you can merge this one, what do you think ?

ulupo:
Perfect!

MonkeyBreaker:
I tried to template the related code. I failed for the simple reason that when calling `get_max_vertex` we need to at least return 2 vertices. I initially though that at the beginning we were returning only one vertex.
Anyway, the commit and the performance improvements LGTM !

I still think that we can continue with the the improvements related to this specific function as you are doing @ulupo !

Also, this code change should be forwarded to upstream `ripser`, what do you think @ulupo ?

We can merge the PR, we know the reason why Python 3.10 fails and unfortunately for the moment we can't do anything.

MonkeyBreaker (approved):

ulupo:
Thanks @MonkeyBreaker !!

ulupo:
I have already notified @ubauer privately!

PR #52, opened 2021-11-15
[TEST] Do not upgrade setuptools in CI

ulupo:


ulupo:
Clearly this doesn't resolve our CI failures for Python 3.10 on Ubuntu.  It's unclear to me what version of `setuptools` is being used here.

ulupo:
Closing as i was mis-diagnosing the reason for the CI failures here.  The real reason is that `scikit-learn` wheels for Python 3.10 were never actually released officially, we had some unofficial ones cached in the runners for a while, but then they disappeared later.

Issue #37, opened 2021-10-31
Always treat bars with infinite death as essential bars


ulupo:
During work on this PR #29, @MonkeyBreaker and I realized that the current core logic in the matrix reduction algorithm makes it so that we break the following promise to the user: that passing a dense matrix with a few `np.inf` will lead to exactly the same output as passing a sparse version where the `np.inf` entries are just missing entries.  This was never noticed at the level of barcodes because a "finite" bar with death equal to infinity looks just the same as an infinite bar with the same birth.  But flag generators don't lie: we now end up telling the user there are never any generators for essential classes if the matrix is dense, no matter what the matrix is or represents.

Because changing this requires potentially delicate changes to the reduction routine, we decided to discuss what to do about this separately, and open a separate PR if necessary after this one is merged.

_Originally posted by @ulupo in https://github.com/giotto-ai/giotto-ph/issues/29#issuecomment-955710360_

ulupo:
A preliminary approach is in https://github.com/ulupo/giotto-ph/pull/1

MonkeyBreaker:
Maybe we could add the diamon example and the output when running in sparse and dense, where we can observe an unexpected difference ?

ulupo:
@MonkeyBreaker indeed. Here goes:

```
import numpy as np
from gph.python import ripser_parallel

diamond_dm = np.array(
    [[0,      1,      np.inf, 1,      1,      1],
     [0,      0,      1,      np.inf, 1,      1],
     [0,      0,      0,      1,      1,      1],
     [0,      0,      0,      0,      1,      1],
     [0,      0,      0,      0,      0,      np.inf],
     [0,      0,      0,      0,      0,      0]],
    dtype=np.float64
)
diamond_dm += diamond_dm.T

res = ripser_parallel(diamond_dm, metric="precomputed", maxdim=2, return_generators=True)
```
yields the correct barcode
```
[array([[ 0.,  1.],
        [ 0.,  1.],
        [ 0.,  1.],
        [ 0.,  1.],
        [ 0.,  1.],
        [ 0., inf]]),
 array([], shape=(0, 2), dtype=float64),
 array([[ 1., inf]])]
```
but the incorrect generators
```
(array([[5, 5, 3],
       [3, 5, 2],
       [2, 5, 1],
       [1, 5, 0],
       [4, 4, 3]], dtype=int64),
[...]

PR #53, opened 2021-11-21
Proposal to deal with explicit infinite edges

ulupo:
Just a way to start the discussion on whether and how to address #37.  I believe this proposal has at least the correct logic, so could serve as an initial reference.

### DIFF
--- a/gph/src/ripser.h
+++ b/gph/src/ripser.h
@@ -94,6 +94,7 @@ typedef uint16_t coefficient_t;
 using barcodes_t = std::vector<value_t>;
 
 const size_t num_coefficient_bits = 8;
+constexpr value_t inf_value = std::numeric_limits<value_t>::infinity();
 
 // 1L on windows is ALWAYS 32 bits, when on unix systems is pointer size
 static const index_t max_simplex_index =
@@ -394,7 +395,7 @@ struct sparse_distance_matrix {
                              index_diameter_t{j, vertex_births[j]});
         return (neighbor != neighbors[i].end() && get_index(*neighbor) == j)
                    ? get_diameter(*neighbor)
-                   : std::numeric_limits<value_t>::infinity();
+                   : inf_value;
     }
 
     size_t size() const { return neighbors.size(); }
@@ -713,7 +714,7 @@ public:
     value_t compute_diameter(const index_t index, const index_t dim) const
     {
         std::vector<index_t> vertices;
-        value_t diam = -std::numeric_limits<value_t>::infinity();
+        value_t diam = -inf_value;
 
         vertices.resize(dim + 1);
         get_simplex_vertices(index, dim, dist.size(), vertices.rbegin());
@@ -980,7 +981,7 @@ public:
         value_t birth, death;
         diameter_index_t birth_vertex;
         for (auto e : edges) {
-            if (get_diameter(e) == std::numeric_limits<value_t>::infinity())
+            if (get_diameter(e) == inf_value)
                 continue;
 
             get_simplex_vertices(get_index(e), 1, n, vertices_of_edge.rbegin());
@@ -1017,8 +1018,7 @@ public:
         for (index_t i = 0; i < n; ++i) {
             if (dset.find(i) == i) {
                 births_and_deaths_by_dim[0].push_back(dset.get_birth(i));
-                births_and_deaths_by_dim[0].push_back(
-                    std::numeric_limits<value_t>::infinity());
+                births_and_deaths_by_dim[0].push_back(inf_value);
                 if (return_flag_persistence_generators) {
                     flag_persistence_generators.essential_0.push_back(
                         dset.get_birth_idx(i));
@@ -1245,7 +1245,7 @@ public:
     edge_t
     get_youngest_edge_simplex(const std::vector<index_t>& vertices_simplex)
     {
-        value_t diam = -std::numeric_limits<value_t>::infinity();
+        value_t diam = -inf_value;
         edge_t edge;
 
         for (index_t i = 0; i < vertices_simplex.size(); ++i) {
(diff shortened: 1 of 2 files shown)


gph/src/ripser.h
```diff
@@ -1564,11 +1587,9 @@ class ripser
                       std::greater<>());
 #endif
             for (size_t i = 0; i < idx_essential; ++i) {
-                if (!std::isinf(essential_pair[i])) {
```
ulupo, line 1567: This is about checking ensuring that births are not infinite ("the bar is never born"), but in this proposal it is not needed because `is_born` makes sure we don't end up in this case.

gph/src/ripser.h
```diff
@@ -991,33 +991,35 @@ class ripser
...
-                death = get_diameter(e);
-                if (death > birth) {
-                    births_and_deaths_by_dim[0].push_back(birth);
-                    births_and_deaths_by_dim[0].push_back(death);
-
-                    if (return_flag_persistence_generators) {
-                        flag_persistence_generators.finite_0.push_back(
-                            {birth_vertex.second, std::max(v1, v2),
-                             std::min(v1, v2)});
+            if (get_diameter(e) != std::numeric_limits<value_t>::infinity()) {
```
ulupo, line 994: If an edge is infinity, it must be skipped (it means it does not exist).

ulupo:
One concern about this PR or similar approaches is that the additional control flow might introduce performance regressions, so I would suggest running extensive profiling before merging if we decide to address #37 at all.

gph/src/ripser.h
MonkeyBreaker: That is why we don't even need to do the link, right ?

MonkeyBreaker: Let me suggest a slightly different code:

```cpp
if (std::isinf(get_diameter(e)))
    continue;
```

And then the rest comes back to the same level of indentation.

BTW, It was my fault, but I think we need to start to use `std::isinf` when comparing to infinity.

gph/src/ripser.h
```diff
@@ -1434,48 +1437,71 @@ class ripser
...
-                            edge_t death_edge =
-                                get_youngest_edge_simplex(vertices_death);
-
-                            finite_generator[location] = {
-                                birth_edge.first, birth_edge.second,
-                                death_edge.first, death_edge.second};
+                        // Infinite death indicates an essential bar in disguise
+                        value_t death = get_diameter(pivot);
+                        is_essential =
+                            (death == std::numeric_limits<value_t>::infinity());
```
MonkeyBreaker, line 1443:
```suggestion
                        is_essential = std::isinf(death);
                 
```

gph/src/ripser.h
```diff
@@ -1434,48 +1437,71 @@ class ripser
...
+                    break;
+                }
+            }
+
+            if (is_essential) {
+                value_t birth = get_diameter(column_to_reduce);
+                /* Infinite birth means the bar is never born and should
+                 * not be included. */
+                bool is_born =
+                    (birth != std::numeric_limits<value_t>::infinity());
```
MonkeyBreaker, line 1501:
```suggestion
                bool is_born = !std::isinf(birth);
                    
```

MonkeyBreaker:
> One concern about this PR or similar approaches is that the additional control flow might introduce performance regressions, so I would suggest running extensive profiling before merging if we decide to address #37 at all.

My intuition after looking at the changes is that the overhead will be not be noticed. A benchmark will confirm this, but I am pretty sure we won't notice a slowdown.

gph/src/ripser.h
ulupo: I considered that option, but didn't implement it in the end because (as far as I understand), `isinf` will also yield `true` for negative infinity, but we don't want to `continue` in that case.  I know it would be very pathological input, but still...

gph/src/ripser.h
ulupo: Same comment as above concerning negative infinity.

gph/src/ripser.h
ulupo: As above.

ulupo:
> A benchmark will confirm this, but I am pretty sure we won't notice a slowdown.

Great!

gph/src/ripser.h
MonkeyBreaker: It is okay, I did not think about this case.
The yes, lets continue with `std::numeric_limits<value_t>::infinity()`.

Then, what do you think of updating the code like that :

```cpp
if (get_diameter(e) == std::numeric_limits<value_t>::infinity())
    continue;
```

ulupo: > That is why we don't even need to do the link, right ?

Right!

> Then, what do you think of updating the code like that :

I'm guessing it reduces indentation and divergence from Ripser/ripser, so why not!

ulupo: Done in f499f17. Maybe something similar can be done in other parts?

MonkeyBreaker:
@ulupo after resolving merging conflicts with latest master I just performed a benchmark:

| dataset | dim | thresh | before timing [s] | after timing  [s] |
|---------|-----|--------|-------------------|-------------------|
| sphere  | 2   | -      | 0.55              | 0.56              |
| dragon  | 1   | -      | 1.66              | 1.70              |
| random  | 7   | -      | 3.53              | 3.5               |
| fractal | 2   | -      | 4.24              | 4.2               |
| o3_1024 | 3   | 1.8    | 1.97              | 1.97              |
| o3_4096 | 3   | 1.4    | 37.5              | 37.4              |
| torus   | 2   | 0.15   | 41.4              | 41.6              |

It seems that the runtime impact is not relevant. All the benchmark were done with `n_threads=1`.

ulupo:
@MonkeyBreaker fantastic about the runtimes!

MonkeyBreaker:
@ulupo thank you for this PR and exploring cases where we had some issues !
I added some minor changes, hope they are good for you.

I will proceed with the Merge of this PR, thank you again for your great work :)

PR #54, opened 2021-11-21
[PY] Compute sparse matrix when input is a point-cloud with a threshold applied

### LINKED ISSUE #55: Incorrect handling of zeros for dense input when a threshold is passed
https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L500-L501 introduces an error: off-diagonal zeros in the distance matrix are treated just like absent edges (which is the same as *infinite* edges) because they are thrown away by `coo_matrix`.

MonkeyBreaker:
This PR adds an optimization for computing the distance matrix from a point cloud directly in Python.

We observed that the computation of a distance matrix for a big point cloud (50k points), is quite expensive.
The upstream C++ `ripser`, when passing a point-cloud with a threshold, computes directly a sparse distance matrix with all distances above the threshold removed.

We reproduce this behavior directly in Python by using a method from `scipy` library.

* [x] Implement the method in Python
    * [x] ~Support for `euclidean` metric~
    * [x] ~Support for `minkowski` metric~
    * [x] ~Support to any other metric ?~
    * [x] Use `sklearn` NearestNeighbors instead of `scipy` approach.
* [x] Add test that compares 2 run on the same dataset, with one using a threshold (enclosing radius) and the other with threshold.
    * [x] ~Test using `euclidean` metric~
    * [x] ~Test using `minkowski` metric~
* [x] Document this behavior somewhere ?
    * [x] : Rework the documentation of X as suggested by @ulupo 
* [x] Sanitize sparse input distance matrix
* [x] Fix #55

### DIFF
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -57,46 +57,44 @@ def _compute_ph_vr_sparse(I, J, V, N, maxHomDim, thresh=-1, coeff=2,
     return ret
 
 
-def _resolve_symmetry_conflicts(coo):
-    """Given a sparse matrix in COO format, filter out any entry at location
-    (i, j) strictly below the diagonal if the entry at (j, i) is also
-    stored. Return row, column and data information for an upper diagonal
-    COO matrix."""
-    _row, _col, _data = coo.row, coo.col, coo.data
+def _sanitize_coo(row, col, data, only_extract_upper=False):
+    """Given a sparse matrix in COO format, either return its upper triangular
+    portion directly, or filter out any entry at location (i, j) strictly below
+    the diagonal if the entry at (j, i) is also stored."""
 
-    below_diag = _col < _row
-    # Check if there is anything below the main diagonal
-    if below_diag.any():
-        # Initialize filtered COO data with information in the upper triangle
-        in_upper_triangle = np.logical_not(below_diag)
-        row = _row[in_upper_triangle]
-        col = _col[in_upper_triangle]
-        data = _data[in_upper_triangle]
+    row_orig, col_orig, data_orig = row, col, data
 
-        # Filter out entries below the diagonal for which entries at
-        # transposed positions are already available
-        upper_triangle_indices = set(zip(row, col))
-        additions = tuple(
-            zip(*((j, i, x) for (i, j, x) in zip(_row[below_diag],
-                                                 _col[below_diag],
-                                                 _data[below_diag])
-                  if (j, i) not in upper_triangle_indices))
-            )
-        # Add surviving entries below the diagonal to final COO data
-        if additions:
-            row_add, col_add, data_add = additions
-            row = np.concatenate([row, row_add])
-            col = np.concatenate([col, col_add])
-            data = np.concatenate([data, data_add])
-
+    # Initialize filtered COO data with information in the upper triangle
+    in_upper_triangle = row_orig <= col_orig
+    row = row_orig[in_upper_triangle]
+    col = col_orig[in_upper_triangle]
+    data = data_orig[in_upper_triangle]
+    if only_extract_upper:
         return row, col, data
-    else:
-        return _row, _col, _data
+
+    below_diag_idxs = np.flatnonzero(np.logical_not(in_upper_triangle))
+    # Check if there is anything below the main diagonal
+    if len(below_diag_idxs):
+        # Only keep entries below the diagonal for which entries at transposed
+        # positions are not available
+        upper_triangle_row_col = set(zip(row, col))
+        additions_idxs = [
+            i for i in below_diag_idxs
+            if (col_orig[i], row_orig[i]) not in upper_triangle_row_col
+            ]
+        # Add surviving entries below the diagonal to final COO data
+        if len(additions_idxs):
+            row = np.concatenate([row, col_orig[additions_idxs]])
+            col = np.concatenate([col, row_orig[additions_idxs]])
+            data = np.concatenate([data, data_orig[additions_idxs]])
+
+    return row, col, data
 
 
 def _collapse_coo(row, col, data, thresh):
     """Run edge collapser on off-diagonal data and then reinsert diagonal
     data."""
+
     diag = row == col
     row_diag, col_diag, data_diag = row[diag], col[diag], data[diag]
     row, col, data = gph_collapser. \
@@ -127,6 +125,7 @@ def _weight_filtration(dist, weights_x, weights_y, p):
     column vector, `weights_y` is a 1D array, `dist` is the original distance
     matrix, and the computations exploit array broadcasting. For sparse data,
     all three are 1D arrays. `p` can only be ``numpy.inf``, ``1``, or ``2``."""
+
     if p == np.inf:
         return np.maximum(dist, np.maximum(weights_x, weights_y))
     elif p == 1:
@@ -471,6 +470,7 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             "`collapse_edges` and `return_generators`cannot both be True."
         )
 
+    is_sparse_matrix_symmetric = None
     if metric == 'precomputed':
         dm = X
     else:
@@ -482,6 +482,7 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                 p = metric_params.get('p', 2)
 
             dm = _pc_to_sparse_dm_with_threshold(X, thresh, p)
+            is_sparse_matrix_symmetric = True
         else:
             dm = pairwise_distances(X, metric=metric, **metric_params)
 
@@ -489,7 +490,11 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
 
     use_sparse_computer = True
     if issparse(dm):
-        row, col, data = _resolve_symmetry_conflicts(dm.tocoo())  # Upper diag
+        coo = dm.tocoo()
+        row, col, data = _sanitize_coo(
+            coo.row, coo.col, coo.data,
+            only_extract_upper=is_sparse_matrix_symmetric
+            )
 
         if weights is not None:
             sparse_kwargs = {
@@ -509,18 +514,19 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             dm = _compute_weights(dm, weights, weight_params, n_points)
 
         compute_enclosing_radius = False
-        if not (dm.diagonal() != 0).any():
+        nonzero_in_diag = (dm.diagonal() != 0).any()
+        if not nonzero_in_diag:
             # Compute ideal threshold only when a distance matrix is passed
             # as input without specifying any threshold
             # We check if any element and if no entries is present in the
-            # diagonal. This allow to have the enclosing radius before
+            # diagonal. This allows to have the enclosing radius before
             # calling collapser if computed
             if thresh == np.inf:
                 thresh = _ideal_thresh(dm, thresh)
                 compute_enclosing_radius = True
 
-        if (dm.diagonal() != 0).any():
-            # Convert to sparse format, because currently that's the only
+        if nonzero_in_diag:
+            # Convert to sparse format, because currently that's the only one
             # one handling nonzero births
             (row, col) = np.triu_indices_from(dm)
             data = dm[(row, col)]
@@ -533,8 +539,10 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
         elif not compute_enclosing_radius:
             # If the user specifies a threshold, we use a sparse
             # representation like Ripser does
-            dm[dm > thresh] = 0.0
-            row, col, data = _resolve_symmetry_conflicts(coo_matrix(dm))
+            row, col = np.nonzero(dm <= thresh)
+            data = dm[(row, col)]
+            row, col, data = _sanitize_coo(row, col, data,
+                                           only_extract_upper=True)
         else:
             use_sparse_computer = False
 
@@ -548,7 +556,8 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             thresh,
             coeff,
             n_threads,
-            return_generators)
+            return_generators
+            )
     else:
         # Only consider strict upper diagonal
         DParam = squareform(dm, checks=False).astype(np.float32)
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -464,7 +464,7 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             if metric == 'euclidean':
                 p = 2
             elif metric == 'minkowski':
-                p = metric_params['p']
+                p = metric_params.get('p', 2)
 
             dm = _pc_to_sparse_dm_with_threshold(X, thresh, p)
         else:
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -267,16 +267,33 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                     metric_params={}, weights=None, weight_params=None,
                     collapse_edges=False, n_threads=1,
                     return_generators=False):
-    """Compute persistence diagrams for X data array.
+    """Compute persistence diagrams from an input dense array or sparse matrix.
 
-    If X is a point cloud, it will be converted to a distance matrix
-    using the chosen metric.
+    If `X` represents a point cloud, a distance matrix will be internally
+    created using the chosen metric and its Vietoris–Rips persistent homology
+    will be computed.
 
     Parameters
     ----------
-    X : ndarray of shape (n_samples, n_features)
-        A numpy array of either data or distance matrix. Can also be a sparse
-        distance matrix of type scipy.sparse
+    X : ndarray or sparse matrix
+        If `metric` is not set to ``"precomputed"``, input data of shape
+        ``(n_samples, n_features)`` representing a point cloud. Otherwise,
+        dense or sparse input data of shape ``(n_samples, n_samples)``
+        representing a distance matrix or adjacency matrix of a weighted
+        undirected graph, with the following conventions:
+            - Diagonal entries indicate vertex weights, i.e. the filtration
+              parameters at which vertices appear.
+            - If `X` is dense, only its upper diagonal portion (including the
+              diagonal) is considered.
+            - If `X` is sparse, it does not need to be upper diagonal or
+              symmetric. If only one of entry (i, j) and (j, i) is stored, its
+              value is taken as the weight of the undirected edge {i, j}. If
+              both are stored, the value in the upper diagonal is taken.
+              Off-diagonal entries which are not explicitly stored are treated
+              as infinite, indicating absent edges.
+            - Entries of `X` should be compatible with a filtration, i.e.
+              the value at index (i, j) should be no smaller than the
+              values at diagonal indices (i, i) and (j, j).
 
     maxdim : int, optional, default: ``1``
         Maximum homology dimension computed. Will compute all dimensions lower
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -249,7 +249,7 @@ def _ideal_thresh(dm, thresh):
     return min([enclosing_radius, thresh])
 
 
-def _pc_to_sparse_dm_with_threshold(X, thresh, algorithm, leaf_size,
+def _pc_to_sparse_dm_with_threshold(X, thresh, nearest_neighbors_params,
                                     metric, metric_params, n_threads):
     """Compute a sparse matrix of pairwise distances between points in a point
     cloud, removing all distances larger than a threshold.
@@ -257,11 +257,10 @@ def _pc_to_sparse_dm_with_threshold(X, thresh, algorithm, leaf_size,
     Return the output as an upper triangular sparse matrix in CSR format."""
 
     neigh = NearestNeighbors(radius=thresh,
-                             algorithm=algorithm,
-                             leaf_size=leaf_size,
                              metric=metric,
                              metric_params=metric_params,
-                             n_jobs=n_threads).fit(X)
+                             n_jobs=n_threads,
+                             **nearest_neighbors_params).fit(X)
     # Upper triangular CSR output
     dm = triu(neigh.radius_neighbors_graph(mode="distance"))
 
@@ -482,10 +481,9 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
     if metric == 'precomputed':
         dm = X
     elif thresh != np.inf:
-        algorithm = nearest_neighbors_params.get("algorithm", "auto")
-        leaf_size = nearest_neighbors_params.get("leaf_size", 30)
         dm = _pc_to_sparse_dm_with_threshold(
-            X, thresh, algorithm, leaf_size, metric, metric_params, n_threads
+            X, thresh, nearest_neighbors_params, metric, metric_params,
+            n_threads
             )
         is_dm_sparse_and_upper_triangular = True
     else:
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -459,7 +459,8 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
     if metric == 'precomputed':
         dm = X
     else:
-        if thresh != np.inf and metric in ['euclidean', 'minkowski']:
+        if not issparse(X) and thresh != np.inf and \
+                metric in ['euclidean', 'minkowski']:
             if metric == 'euclidean':
                 p = 2
             elif metric == 'minkowski':
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -470,7 +470,8 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             "`collapse_edges` and `return_generators`cannot both be True."
         )
 
-    is_sparse_matrix_symmetric = None
+    use_sparse_computer = True
+    is_dm_sparse_and_symmetric = False
     if metric == 'precomputed':
         dm = X
     else:
@@ -482,18 +483,17 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
                 p = metric_params.get('p', 2)
 
             dm = _pc_to_sparse_dm_with_threshold(X, thresh, p)
-            is_sparse_matrix_symmetric = True
+            is_dm_sparse_and_symmetric = True
         else:
             dm = pairwise_distances(X, metric=metric, **metric_params)
 
     n_points = max(dm.shape)
 
-    use_sparse_computer = True
     if issparse(dm):
         coo = dm.tocoo()
         row, col, data = _sanitize_coo(
             coo.row, coo.col, coo.data,
-            only_extract_upper=is_sparse_matrix_symmetric
+            only_extract_upper=is_dm_sparse_and_symmetric
             )
 
         if weights is not None:
--- a/gph/python/ripser_interface.py
+++ b/gph/python/ripser_interface.py
@@ -304,8 +304,10 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
         than or equal to this value.
 
     thresh : float, optional, default: ``numpy.inf``
-        Maximum distances considered when constructing filtration. If
-        ``numpy.inf``, compute the entire filtration.
+        Maximum value of the Vietoris–Rips filtration parameter. Points whose
+        distance is greater than this value will never be connected by an edge.
+        If ``numpy.inf``, compute the entire filtration. Otherwise, and if
+        `metric` is not ``"precomputed"``, see `nearest_neighbors_params`.
 
     coeff : int prime, optional, default: ``2``
         Compute homology with coefficients in the prime field Z/pZ for p=coeff.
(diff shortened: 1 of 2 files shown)


gph/python/ripser_interface.py
```diff
@@ -447,7 +459,16 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
     if metric == 'precomputed':
         dm = X
     else:
-        dm = pairwise_distances(X, metric=metric, **metric_params)
+        if not issparse(X) and thresh != np.inf and \
+                metric in ['euclidean', 'minkowski']:
```
MonkeyBreaker, line 463: @ulupo this is a draft, I do not think that this solution is easy to scale up if we need to support more metrics.

The other issue is that `cKDTree` does not support sparse input. I choose to not go with the optimized solution if the input is sparse.

Let me know if you have a different solution/approach.

ulupo: This looks reasonable for now.

gph/python/ripser_interface.py
```diff
@@ -447,7 +459,16 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
...
+            if metric == 'euclidean':
+                p = 2
+            elif metric == 'minkowski':
+                p = metric_params['p']
```
ulupo, line 467: I think this should be `p = metric_params.get('p', 2)` because `minkowski` has default `2` e.g. in `pdist`.

ulupo (commented):
I think this PR is a good opportunity to also improve on a couple of highly related things:
  1. Docstrings: there are a few rules about how the input `X` is processed that are explicitly stated in `giotto-tda` and should also be explicitly stated here because they are (I think) the same.  Here we should be able to import the content of https://github.com/giotto-ai/giotto-tda/blob/a0f4cd4a5cf179b25c9ab7e5aee5ca889b4fc183/gtda/homology/simplicial.py#L204-L220 into the description of the `X` parameter.
  2. Use of `_resolve_symmetry_conflicts`: the only use case for this function as is written now is when `metric='precomputed'` and the input is sparse.  In the current code, it is executed in https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L501 unnecessarily because the rule for dense matrices is that the lower diagonal is ignored completely, so we should just fetch the upper diagonal `row`, `col` and `data` values.  In the same way, https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L456 now gets run even if `_pc_to_dm_with_threshold` is applied.  But `_pc_to_dm_with_threshold` returns a symmetric matrix, so again `_resolve_symmetry_conflicts` will do useless work in this case.  A possible solution is to have an additional kwarg `check` (default `True`) to `_resolve_symmetry_conflicts`: when `False`, just the upper diagonal information is returned.

gph/python/ripser_interface.py
```diff
@@ -251,6 +252,17 @@ def _ideal_thresh(dm, thresh):
     return min([enclosing_radius, thresh])
 
 
+def _pc_to_dm_with_threshold(pc, thresh, p):
```
ulupo, line 255:
```suggestion
def _pc_to_sparse_dm_with_threshold(pc, thresh, p):
```

ulupo (commented):
I suggest this name change.

gph/python/ripser_interface.py
MonkeyBreaker: You are right !

MonkeyBreaker: Done

MonkeyBreaker:
> Docstrings: there are a few rules about how the input X is processed that are explicitly stated in giotto-tda and should also be explicitly stated here because they are (I think) the same. Here we should be able to import the content of https://github.com/giotto-ai/giotto-tda/blob/a0f4cd4a5cf179b25c9ab7e5aee5ca889b4fc183/gtda/homology/simplicial.py#L204-L220 into the description of the X parameter.

If I understood you right, you would like to replace the current docstring for parameter X to the following:

```
            X : ndarray or list of length n_samples
            Input data representing a collection of point clouds if `metric`
            was not set to ``"precomputed"``, and of distance matrices or
            adjacency matrices of weighted undirected graphs otherwise. Can be
            either a 3D ndarray whose zeroth dimension has size ``n_samples``,
            or a list containing ``n_samples`` 2D ndarrays/sparse matrices.
            Point cloud arrays have shape ``(n_points, n_dimensions)``, and if
            `X` is a list these shapes can vary between point clouds. If
            `metric` was set to ``"precomputed"``, then:
                - Diagonal entries indicate vertex weights, i.e. the filtration
                  parameters at which vertices appear.
                - If entries of `X` are dense, only their upper diagonal
                  portions (including the diagonal) are considered.
                - If entries of `X` are sparse, they do not need to be upper
                  diagonal or symmetric. If only one of entry (i, j) and (j, i)
                  is stored, its value is taken as the weight of the undirected
                  edge {i, j}. If both are stored, the value in the upper
                  diagonal is taken. Off-diagonal entries which are not
                  explicitly stored are treated as infinite, indicating absent
                  edges.
                - Entries of `X` should be compatible with a filtration, i.e.
                  the value at index (i, j) should be no smaller than the
                  values at diagonal indices (i, i) and (j, j).
 ```
 
 Is that right ?

**Comment by ulupo** [comment]
> If I understood you right, you would like to replace the current docstring for parameter X to the following:

Almost. But some things need to be changed because I pulled that out of `VietorisRipsPersistence` which processes *collections* of distance matrices/point clouds.  I'll adapt that text ASAP, please standby :).

**Comment by MonkeyBreaker** [comment]
> Use of _resolve_symmetry_conflicts: the only use case for this function as is written now is when metric='precomputed' and the input is sparse. In the current code, it is executed in
giotto-ph/gph/python/ripser_interface.py Line 501 in 2b3d676
 row, col, data = _resolve_symmetry_conflicts(coo_matrix(dm)) 
unnecessarily because the rule for dense matrices is that the lower diagonal is ignored completely, so we should just fetch the upper diagonal row, col and data values. In the same way,
giotto-ph/gph/python/ripser_interface.py Line 456 in 2b3d676
 row, col, data = _resolve_symmetry_conflicts(dm.tocoo())  # Upper diag 
now gets run even if _pc_to_dm_with_threshold is applied. But _pc_to_dm_with_threshold returns a symmetric matrix, so again _resolve_symmetry_conflicts will do useless work in this case. A possible solution is to have an additional kwarg check (default True) to _resolve_symmetry_conflicts: when False, just the upper diagonal information is returned.

I am not sure it is true, if I follow correctly the code, we call 2 times `_resolve_symmetry_conflicts`:

1. https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L456
2. https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L501

The first one is computed in all cases if the input is sparse. I think that here you are right, we should not compute this if we have a sparse input because we called `_pc_to_sparse_dm_with_threshold`. But as you point out, I think that `_pc_to_sparse_dm_with_threshold` should only return the upper diagonal part, if you can confirm this, then I can proceed to the necessary changes inside of the function itself.

For the second case that `_resolve_symmetry_conflicts` is called, it will only be when the input a distance matrix (`metric='precomputed'`) and a threshold is provided. But on your message I am not sure I understand the following remark:

>  In the current code, it is executed in ... unnecessarily because the rule for dense matrices is that the lower diagonal is ignored completely, so we should just fetch the upper diagonal row, col and data values.

So instead of calling `_resolve_symmetry_conflicts`, we should only remove all entries below the diagonal, right ?

**Comment by MonkeyBreaker** [comment]
> Almost. But some things need to be changed because I pulled that out of VietorisRipsPersistence which processes collections of distance matrices/point clouds. I'll adapt that text ASAP, please standby :).

Sure, no stress, feel free to directly push your suggestion into the PR :)

**Comment by ulupo** [comment]
> But on your message I am not sure I understand the following remark

This remark is about the fact that https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/python/ripser_interface.py#L501 is only called when the user passed dense input.

I don't think the best approach is

> _pc_to_sparse_dm_with_threshold should only return the upper diagonal part

I think the best approach is

>  possible solution is to have an additional kwarg check (default True) to _resolve_symmetry_conflicts: when False, just the upper diagonal information is returned.

Because `_resolve_symmetry_conflicts` is already about making decisions about lower vs upper diagonals, so it makes sense to me that it should take full responsibility of it in all cases.

**Comment by ulupo** [comment]
@MonkeyBreaker in view of #55, I think we can think about fixing that issue here as suggested there, what do you think?

**Comment by MonkeyBreaker** [comment]
> @MonkeyBreaker in view of #55, I think we can think about fixing that issue here, what do you think?

Sure, it makes sense because we are reworking the processing of the input :) !

> because _resolve_symmetry_conflicts is already about making decisions about lower vs upper diagonals, so it makes sense to me that it should take full responsibility of it in all cases, if possible.

I understand now, I think the confusion for me is that `_resolve_symmetry_conflicts` does 2 jobs, resolving symmetry issues and returning the upper diagonal part only. Why not instead create another function that from a sparse input, only returns the upper diagonal part ? 

This allows to call something like `_extract_upper_diagonal()` when we only need to do this step and call `_extract_upper_diagonal()` from `_resolve_symmetry_conflicts` directly. This allows to split the job done by function. And we won't need to add a kwarg parameter.

**Comment by ulupo** [comment]
I still prefer a solution with a kwarg, as follows or similar:

```py

def _resolve_symmetry_conflicts(coo, check=True):
    """Given a sparse matrix in COO format, filter out any entry at location
    (i, j) strictly below the diagonal if the entry at (j, i) is also
    stored. Return row, column and data information for an upper diagonal
    COO matrix."""
    _row, _col, _data = coo.row, coo.col, coo.data
    in_upper_triangle = _row <= _col
    if check:
        # Check if there is anything below the main diagonal
        if in_upper_triangle.all():
            # Initialize filtered COO data with information in the upper triangle
            row = _row[in_upper_triangle]
            col = _col[in_upper_triangle]
            data = _data[in_upper_triangle]
            # Filter out entries below the diagonal for which entries at
            # transposed positions are already available
            below_diag = np.logical_not(in_upper_triangle)
            upper_triangle_indices = set(zip(row, col))
            additions = tuple(
                zip(*((j, i, x) for (i, j, x) in zip(_row[below_diag],
                                                     _col[below_diag],
                                                     _data[below_diag])
                      if (j, i) not in upper_triangle_indices))
                )
            # Add surviving entries below the diagonal to final COO data
            if additions:
                row_add, col_add, data_add = additions
                row = np.concatenate([row, row_add])
                col = np.concatenate([col, col_add])
                data = np.concatenate([data, data_add])
            return row, col, data
        else:
            return _row, _col, _data
    else:
        row, col = _row[in_upper_triangle], _col[in_upper_triangle]
        data = _data[in_upper_triangle]
        return row, col, data
```

MonkeyBreaker:
> Having to make a decision about which function to use is in my view more overhead.

I understand but then we need to have a function name that conveys the purpose.

> Maybe something like _resolve_symmetry (so we don't suggest too strongly that there should be conflicts)?

I was more thinking something like `_extract_upper(X, resolve_symmetry_conflicts=False)`. Then the primary purpose would be to extract the upper part. And Additionally it would be able to resolve symmetry conflicts. What do you think ?

ulupo:
> This allows to call something like _extract_upper_diagonal() when we only need to do this step and call _extract_upper_diagonal() from _resolve_symmetry_conflicts directly. This allows to split the job done by function. And we won't need to add a kwarg parameter.

I'm guessing you are saying that if we pass `_resolve_symmetry_conflicts` something already processed by `_extract_upper_diagonal` then we it will just trivially return the initial COO data, and we can exploit this?

MonkeyBreaker:
> I'm guessing you are saying that if we pass to _resolve_symmetry_conflicts something already processed by _extract_upper_diagonal then we it will just trivially return the initial COO data, and we can exploit this?

No I think that for `_resolve_symmetry_conflicts`, after looking at the computation, it needs to know the values below the diagonal.

In an ideal case, yes, we could do both steps separately. But specifically for `_resolve_symmetry_conflicts`, it needs to know the values below the diagonal, therefore we cannot pass it a upper diagonal matrix only.

My previous remark won't apply to `_resolve_symmetry_conflicts`, I will cross-out this part to make things more clear.

ulupo:
To reply to

> I was more thinking something like _extract_upper(X, resolve_symmetry_conflicts=False). Then the primary purpose would be to extract the upper part. And Additionally it would be able to resolve symmetry conflicts. What do you think ?

I don't think this helps think about things clearly.  The main purpose is to "sanitize" the sparse input.  Now, "sanitizing" means:
1. extracting the upper diagonal if that's all there is
2. "resolving conflicts" if there's stuff both above and below

The rest is just performance considerations.  If we already know that there's stuff above and below (case 2) but no possibility of conflicts, we should shortcut the logic.  This is the purpose of the extra `kwarg`.  I agree that this can be achieved by passing `_resolve_symmetry_conflicts` something that has nothing below the diagonal, but this does not feel like it conveys the purpose as clearly to me.

ulupo:
> No I think that for _resolve_symmetry_conflicts, after looking at the computation, it needs to know the values below the diagonal.

If there's nothing below the diagona, `_resolve_symmetry_conflicts` just returns the input unchanged.

MonkeyBreaker:
> The rest is just performance considerations. If we already know that there's stuff above and below (case 2) but no possibility of conflicts, we should shortcut the logic for performance reasons. This is the purpose of the extra kwarg. I agree that this can be achieved by passing _resolve_symmetry_conflicts something that has nothing below the diagonal, but this does not feel like it conveys the purpose as clearly to me.

XD, I think we won't agree. But I think it is subjective, if I summarize:

* You prefer something along ` _resolve_symmetry_conflicts(X, only_ret_upper_diag)`
* I prefer something along `_extract_upper(X, _resolve_symmetry_conflicts)`

In the end as you wrote:

>  The main purpose is to "sanitize" the sparse input. Now, "sanitizing" means

Then, maybe something more like:

```python
def _sanetize_sparse_input(X, only_extract_upper)
```

Something like that.
I prefer if you choose here, because this is a code you wrote originally.

ulupo:
Thanks for the summary!  Yes I agree -- and just FYI, the reason I'd prefer to avoid `_extract_upper` is that, if I understand your proposal, this function might end up extracting and returning values which came from *below* the diagonal.  This I find confusing, and is the reason for my opposition.

> Then, maybe something more like:
> ```py
> def _sanitize_sparse_input(X, only_extract_upper)
> ```

This is a great suggestion!  Maybe `_sanitize_sparse_dm` because it sanitizes a sparse matrix rather than the "raw" information contained in the `row`, `col` and `data` objects.

MonkeyBreaker:
>  I'd prefer to avoid _extract_upper is that, if I understand your proposal, this function might end up extracting and returning values which came from below the diagonal

Oh crap, right, If a value is missing in the upper part but present in the lower part, it would use the value from the lower part. Now I got it !

> OR we make it handle that "raw" information instead! This way we can use this same function to deal with #55 too!

Well if it keeps the function simple to understand, sure !

ulupo:
> Oh crap, right, If a value is missing in the upper part but present in the lower part, it would use the value from the lower part. 

Indeed! :)

> Well if it keeps the function simple to understand, sure !

I'll give a proposal.

ulupo:
@MonkeyBreaker fixed the docstrings!

ulupo:
@MonkeyBreaker I pushed 24f2715 to show you give a proposal.  Feel free to review!

Once this is done, it would be good to add a regression test for #55.  I can do this.

ulupo:
From the [SciPy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html):

> For large dimensions (20 is already large) do not expect this to run significantly faster than brute force. High-dimensional nearest-neighbor queries are a substantial open problem in computer science.

So this solution might introduce a performance regression when `X.shape[1]` is large.  If `X.shape[0]` is very large we have no choice for memory reasons, but perhaps we should exclude this method when `X.shape[1]` is large (say > 20) but `X.shape[0]` is small enough that the dense distance matrix will take no more than e.g. a few hundred MB in memory.

gph/python/ripser_interface.py
```diff
@@ -56,46 +57,44 @@ def _compute_ph_vr_sparse(I, J, V, N, maxHomDim, thresh=-1, coeff=2,
...
+def _sanitize_coo(row, col, data, only_extract_upper=False):
+    """Given a sparse matrix in COO format, either return its upper triangular
+    portion directly, or filter out any entry at location (i, j) strictly below
+    the diagonal if the entry at (j, i) is also stored."""
 
-    below_diag = _col < _row
+    row_orig, col_orig, data_orig = row, col, data
+
+    # Initialize filtered COO data with information in the upper triangle
+    in_upper_triangle = row_orig <= col_orig
```
MonkeyBreaker, line 68: Now we also retrieve the diagonal, is that right ?

ulupo: I thought this was the case before too?

MonkeyBreaker: You are right, I did not see that row and column were exchanged, sorry

ulupo: No problem!

gph/python/ripser_interface.py
```diff
@@ -444,16 +470,31 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             "`collapse_edges` and `return_generators`cannot both be True."
         )
 
+    is_sparse_matrix_symmetric = None
```
MonkeyBreaker, line 473:
```suggestion
    is_sparse_matrix_symmetric = False
```

Is there a reason why initialize with `None` ?

ulupo: It's not a strong reason, but it is to indicate to the reader that this is an optional thing, i.e. it does not apply to the dense cases obviously.

MonkeyBreaker: Okay, I find it weird, but if this is a convention, then lets keep it with `None`

ulupo: I could maybe name it `is_dm_sparse_and_symmetric` so I don't mean "is the sparse matrix symmetric" but "is the distance matrix sparse and symmetric", and this variable makes sense in all cases.  What do you think?

MonkeyBreaker: Definitely, the name is better.
And it makes more sense.

BTW, could we move some  `use_sparse_computer` declaration at the beginning after the check ?
So we have the declaration in the same part of the code.

ulupo: I'll change the name, but could you indicate which line you'd like `use_sparse_computer` in?

MonkeyBreaker:
> So this solution might introduce a performance regression when X.shape[1] is large. If X.shape[0] is very large we have no choice for memory reasons, but perhaps we should exclude this method when X.shape[1] is large (say > 20) but X.shape[0] is small enough that the dense distance matrix will take no more than e.g. one or two hundred MB in memory (e.g. if X.shape[0] is 5000 them the dense dm would take 200MB).

Hum, this is an interesting point.
I would prefer to not have a custom threshold, but to let the user the possibility to bypass this.

I mean we should do some testing and see how both methods behave when increasing `X.shape[0]` or `X.shape[1]` independently. We already know that the more we increase `X.shape[0]` the better it is to use the new optimization.
We will need to benchmark for `X.shape[1]`, I can maybe do this tonight, for a dataset of 100 points, increasing the dimensions of the points.

But I do not think that some results on yours or mine computer could provide a specific threshold. I think that it would be better to define a way to bypass this with a `kwarg` or something similar.

@ulupo What do you think ?

gph/python/ripser_interface.py
MonkeyBreaker: Just below or before `is_dm_sparse_and_symmetric` declaration.
I think it will be better for reader to centralize when it make sense variable declaration.
And also in my opinion to make code more readable.

MonkeyBreaker:
Also in after we find a solution for your remark about the increasing dimensionality of the points..

2 Reaming tasks to do for this PR:

1. @ulupo, Do you think we should support more metrics for this first version ?
2. Should we document this behavior somewhere specific, by behavior I mean to use the new method for dense input, using a threshold for specific metric ?

gph/python/ripser_interface.py
ulupo: Ok!

ulupo:
> 2 Reaming tasks to do for merging this PR:
>    1. @ulupo, Do you think we should support more metrics for this first version ?
>    2. Should we document this behavior somewhere specific, by behavior I mean to use the new method for dense input, using a threshold for specific metric ?

1. I don't know how we could achieve this without bringing back a full C++ solution like Ripser/ripser.
2. I don't think we should add anything to the docstring, because how we compute distance matrices should not be relevant to the user.

MonkeyBreaker:
Perfect, then if we can address the increasing dimensionality of the points issue.
We should be able to merge the PR !

gph/python/ripser_interface.py
ulupo: Both changes done in 34fb487. I let you resolve.

ulupo:
>  I would prefer to not have a custom threshold, but to let the user the possibility to bypass this.

Let me try to explain how I see this issue:

All we really owe to the users is to offer a solution that does not fill up their RAM when this could have be avoided.  The rest (including possible performance improvements) is a bonus but we have no duty towards it now.  In other words, I think we can take the view that the dense (`pdist`) + filtering approach remains the preferred approach (even if one could do better) *unless* `X.shape[0]` is so high that an unacceptable amount of RAM is used.  I agree that we should ultimately let the user override things if they know what they are doing, but we still have to offer a reasonable default even if it's very "unadventurous".  So I propose:
  1. That by default we decide whether to use `KDTree` or not only according to a simple cut-off on `X.shape[0]` -- we decide in this PR how much RAM would be unacceptable for a dense `dm` and choose this cut-off accordingly.
  2. That we still allow advanced users to override the default rule with an optional kwarg.

The optional kwarg could be called e.g. `use_kdtree` and it could have three values: `"auto"` (default, as above), `True` and `False`.

Relevant to this discussion: https://en.wikipedia.org/wiki/K-d_tree#Degradation_in_performance_with_high-dimensional_data

MonkeyBreaker:
> All we really owe to the users is to offer a solution that does not fill up their RAM when this could have been avoided. The rest (including possible performance improvements) is a bonus but we have no duty towards it in this PR. In other words, I think we can take the view that the dense (pdist) + filtering approach remains the preferred approach (even if one could do better) unless X.shape[0] is so high that an unacceptable amount of RAM is used.

You are right, we need to address a Memory problem and not a runtime issue. To let a trace here about the example we discussed internally, the dataset `torus`, a point-cloud with 50k points. When computing the distance matrix of `torus` we need 20GB of memory (20GB because we use 64-bit floats). With the new approach, the sparse distance matrix, using a threshold of `0.15`, it only needs around 100MB of memory.

> That by default we decide whether to use KDTree or not only according to a simple cut-off on X.shape[0] -- we decide in this PR how much RAM would be unacceptable for a dense dm and choose this cut-off accordingly.

I think this is hard. If the user does not provide a threshold, using `KDTree` approach won't consume less memory. To keep this easy for the moment, we go for the `KDTree` if user inputs a point-cloud with a threshold, otherwise we go with the "standard" computation of the dense distance matrix.

> That we still allow advanced users to override the default rule with an optional kwarg.

I don't know, the more I think about this, the more I think that this should be an internal behavior.
We implement this feature to reduce the memory consumed for the computation of the distance matrix when a threshold is used.
It would become really specific if user would start to bypass this computation because it is faster to compute in a standard way. But because they provided a threshold, the algorithm will still use a sparse representation and therefore a performance penalty will happen during PH computation.

We should keep it as is.

Do you agree @ulupo , or you think it is still a good idea to let the user the possibility to choose ?

ulupo:
> I think this is hard. If the user does not provide a threshold, using KDTree approach won't consume less memory. To keep this easy for the moment, we go for the KDTree if user inputs a point-cloud with a threshold, otherwise we go with the "standard" computation of the dense distance matrix.

Of course this is what I actually meant.  Thanks for making it clearer!

> Do you agree @ulupo , or you think it is still a good idea to let the user the possibility to choose ?

Well, I'm not sure.  If my data is 50k points and it's in 1000 dimensions, KDTree might be incredibly slow as discussed in https://en.wikipedia.org/wiki/K-d_tree#Degradation_in_performance_with_high-dimensional_data.  So if I know what I am doing, I might prefer to use 20GB willingly.

MonkeyBreaker:
> Well, I'm not sure. If my data is 50k points and it's in 1000 dimensions, KDTree might be incredibly slow as discussed in https://en.wikipedia.org/wiki/K-d_tree#Degradation_in_performance_with_high-dimensional_data. So if I know what I am doing, I might prefer to use 20GB willingly.

Right, I did not even think about this case. In this case, yes we should add as you suggested a `kwarg`. I like the name `use_kdtree`, we should explain in the docstring the reason of this argument. Do you think we could provide a rule of thumb, or we say that only advanced user shall be concerned by this argument ?

ulupo:
> Do you think we could provide a rule of thumb, or we say that only advanced user shall be concerned by this argument ?

I think that would be too hard. It's enough to make the readers aware in my opinion.

ulupo:
> Right, I did not even think about this case. In this case, yes we should add as you suggested a kwarg. I like the name use_kdtree, we should explain in the docstring the reason of this argument

Shall I proceed with these changes?  I'll ask for a review once I'm done.

MonkeyBreaker:
> Shall I proceed with these changes? I'll ask for a review once I'm done.

With pleasure !

ulupo:
I pushed an approach based on `scikit-learn`'s `NearestNeighbors` class and should offer the ultimate amount of flexibility.  @MonkeyBreaker the logic for the automatic choice of algorithm is in these lines:
https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/neighbors/_base.py#L522-L543

MonkeyBreaker:
> I believe this shows that "brute" (compute the full distance matrix) is always chosen if the number of dimensions is larger than 15, and that kDTree is preferred otherwise. For torus (4 < 15 dimensions), BallTree is faster, but I think we should stick with the sklearn defaults, what do you think?

[Documentation](https://scikit-learn.org/stable/modules/neighbors.html#choice-of-nearest-neighbors-algorithm) has all details about when each algorithm is selected.

I think as you that we should stick with `sklearn` defaults. It will much easier for us to maintain.

> I realize that some of this contradicts my previous opinion that we should always avoid the dense computation by default if too much RAM is used, but at least with this solution we gain some tight integration with sklearn, full coverage of all metrics, and some degree of reasonable automation at least for low dimensions. We could always add a custom warning if both shape[0] is large and shape[1] is large, what do you think @MonkeyBreaker ?

I think that with the documentation you did in the docstring it is good enough. I don't think we can have a perfect solution but right now I think that we are good !

Let me try to benchmark current defaults to compare how better we are compared to originally.

ulupo (commented):
I missed this in #50 but `max_coeff_supported` should probably be elevated to a global variable defined somewhere at the top of the file. @MonkeyBreaker I can make this small change in this PR if OK with you?

MonkeyBreaker:
> I missed this in #50 but max_coeff_supported should probably be elevated to a global variable defined somewhere at the top of the file. @MonkeyBreaker I can make this small change in this PR if OK with you?

Sure, if you put it as a global variable, could you uppercase the entire variable ? I think it is a convention for Python

ulupo:
> could you uppercase the entire variable
Yup, I was going to do it like that.

gph/python/test/test_ripser.py
```diff
@@ -397,9 +396,8 @@ def test_optimized_distance_matrix(metric):
     maxdim = 2
     enclosing_radius = 0.8884447324918705
 
-    dgms = ripser(X, metric=metric,  maxdim=maxdim)["dgms"]
-    dgms_thresh = ripser(X, metric=metric, maxdim=maxdim,
-                         thresh=enclosing_radius)["dgms"]
+    dgms = ripser(X,  maxdim=maxdim)["dgms"]
```
ulupo, line 399:
```suggestion
    dgms = ripser(X, maxdim=maxdim)["dgms"]
```

MonkeyBreaker:
@ulupo, any important points to resolve that I can help in order to merge this new improvement :smile: !

Btw, 2 things I verified, when using Torus, threshold 0.15 and maxdim = 2:

* Memory peak now : 0.4GB
* Memory peak with `scipy` : 0.7GB
* Memory peak originally : 12GB

MonkeyBreaker:
On my side, if you address my comment on `triu` comment that I think it is not right.
Then it all looks good to me !

ulupo:
> On my side, if you address my comment on triu comment that I think it is not right.

@MonkeyBreaker I'm sorry, I don't understand what comment this is.

MonkeyBreaker:
> @MonkeyBreaker I'm sorry, I don't understand what comment this is.

Sorry, again, I forgot to publish them ...

gph/python/ripser_interface.py
```diff
@@ -252,6 +253,24 @@ def _ideal_thresh(dm, thresh):
...
+    cloud, removing all distances larger than a threshold.
+
+    Return the output as an upper triangular sparse matrix in CSR format."""
+
+    neigh = NearestNeighbors(radius=thresh,
+                             metric=metric,
+                             metric_params=metric_params,
+                             n_jobs=n_threads,
+                             **nearest_neighbors_params).fit(X)
+    # Upper triangular CSR output
```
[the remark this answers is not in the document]
ulupo: Ah I see!  Well, COO is actually better for what we want to do later (I think), so I can just fix the docs. What do you think?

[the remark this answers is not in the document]
MonkeyBreaker: Sure, fixing the docs seems the good idea (I am not sure if we should also explicitely specify `format='coo'`) !
I think too that we should keep using COO whenever we can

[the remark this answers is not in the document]
ulupo: I like being explicit so passed `format="coo"`.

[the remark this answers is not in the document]
MonkeyBreaker: Perfect, I resolve, I do a last review

MonkeyBreaker:
LGTM !
Let's go for the merge :)

I generated the documentation on my machine, looks greats :+1:

PR #58, opened 2021-12-16
[DOCS] Specify that parallel computations are ony available in dims >= 1

### LINKED ISSUE #57: [DOC] Dimension 0 is computed sequentially
It is maybe clear for library developers that dimension 0 is computed sequentially, where all dimensions != 0 are computed in parallel.

This behavior is mentioned in our paper, but it is mentioned nowhere in the docstring of the library.
We should add a mention to this, because users could be mislead that all the Persistent Homology computation is done in parallel.

ulupo:
Fixes #57


MonkeyBreaker (approved):

MonkeyBreaker:
Perfect, thank you so much @ulupo !

PR #59, opened 2021-12-26
Prepare v0.2.1

ulupo:
Bump version and add release notes for 0.2.1.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,3 +1,35 @@
+Release 0.2.1
+=============
+
+Minor release bringing bug fixes and performance improvements.
+
+Major Features and Improvements
+-------------------------------
+
+- Miscellaneous improvements including verification that ``coeff`` is a prime, improvements to the Python bindings (see "Backwards-Incompatible Changes" below), updated C++ preprocessor directives and code style fixes (`#50 <https://github.com/giotto-ai/giotto-ph/pull/50>`_).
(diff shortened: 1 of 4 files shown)


RELEASE.rst
```diff
@@ -1,3 +1,35 @@
...
+
+Bug Fixes
+---------
+
+- A bug in computing sparse thresholded distance matrices has been fixed (`#54 <https://github.com/giotto-ai/giotto-ph/pull/54>`_).
+
+Backwards-Incompatible Changes
+------------------------------
+
+- Barcodes are now returned as (lists of) arrays of dtype ``numpy.float32`` instead of ``numpy.float64``, since single-precision floats are used internally by the C++ backend (`#50 <https://github.com/giotto-ai/giotto-ph/pull/50>`_).
```
MonkeyBreaker, line 22: At first when reading this, I though that we changed the return data structure. Where in reality we only changed the type of the values themself, from `np.float64` to `np.float32`.

I don't know if it is me, or maybe change slightly the wording, what do you think @ulupo?

MonkeyBreaker:
Thank you @ulupo for preparing the release !
Only I minor comment that can be discarded in my opinion.

Great work !

RELEASE.rst
ulupo: I think that since I say

> instead of ``numpy.float64``

the meaning is not ambiguous.

PR #61, opened 2022-01-09
Extend dense matrix backend to allow for nonzero vertex weights

MonkeyBreaker:
This PR introduce support in the C++ back-end of non-zero filtration values in the diagonal during when representing the distance matrix as dense.

# Todo

* [x] : Update C++ `compressed_distance_matrix` to support non-zero diagonals
* [x] : Update Python computation of the vector-form distance conversion from the square distance-matrix.
* [x] : Update when using weights to use a dense representation


MonkeyBreaker:
On standard datasets that use a dense representation during computation, I measured the following timings :

| dataset | dim | thresh | before timing [s] (MB) | after timing  [s] (MB) |
|---------|-----|--------|------------------------|------------------------|
| sphere  | 2   | -      | 0.55 (154.4)           | 0.54 (154.7)           |
| dragon  | 1   | -      | 1.64 (280)             | 1.63 (280)             |
| random  | 7   | -      | 3.5 (210)              | 3.41 (210)             |
| fractal | 2   | -      | 4.1 (104.4)            | 4 (104.5)              |

No performance regression can be observed from standard datasets with the updated backend.

ulupo:
Thanks @MonkeyBreaker!  Concerning

> No performance regression can be observed from standard datasets with the updated backend.

Do we agree that this is not surprising?  In my understanding, in these cases we should follow exactly the same paths as before.  What I'm curious about is how much faster this solution is in the case of interest of a dense matrix with no infinities and no zeros along the diagonal, versus the current solution of converting it to sparse format and then using the sparse computer.

MonkeyBreaker:
> Do we agree that this is not surprising? In my understanding, in these cases we should follow exactly the same paths as before. What I'm curious about is how much faster this solution is in the case of interest of a dense matrix with no infinities and no zeros along the diagonal, versus the current solution of converting it to sparse format and then using the sparse computer.

Definitely, I didn't think that on the standard cases it should change a lot. I was interested if the memory consumption is much higher because now we store a slightly bigger distance matrix. And also about runtime, removing the condition `i==j ? 0 : ...`, I was interested if this could produce observable differences in runtime.

> To produce these cases, one could use the same datasets as in your benchmarks but also use weights.

I will look for adding the weights on 4 datasets and perform a benchmark.

ulupo:
> I was interested if the memory consumption is much higher because now we store a slightly bigger distance matrix.

In your benchmarks though, the distance matrices to be passed are literally identical.  We passed dense before (because no threshold) and we still should pass dense after this PR.

The way I see it, this PR cannot even in principle lead to performance regressions or higher memory consumption.  It should only allow for us to avoid the hack of converting distance matrices which truly should be dense (because no edge is removed) into sparse ones just because there are non-zero diagonal values.  And there is surely no way that using the sparse implementation for a matrix that is actually fully filled can be faster than a dense implementation (notice that for dimensions 1 and above, the diagonal even becomes irrelevant).

MonkeyBreaker:
> In your benchmarks though, the distance matrices to be passed should literally be identical before and after this PR.

They are not exactly the same in the C++, now the `compressed_distance_matrix` contains the diagonal, where previously it didn't.
You can see in the Python interface, I did the following change:

```python
DParam = squareform(dm, checks=False).astype(np.float32)
```
To
```python
DParam = dm[np.triu_indices(dm.shape[0])].astype(np.float32)
```

The new version produces a longer vector containing the diagonal.
But as we can see on the measurement the memory impact is not observable.

> The way I see it, this PR cannot even in principle lead to performance regressions or higher memory consumption.

Because I changed `compressed_distance_matrix`, I wanted to be sure that on standard case no performance regression was present.

MonkeyBreaker:
Benchmark using `weights` with the following parameters:

* `weights='DTM'`
* `weight_params={'p'=1}`

| dataset | dim | thresh | before timing [s] | after timing  [s] | factor |
|---------|-----|--------|-------------------|-------------------|--------|
| sphere  | 2   | -      | 0.58              | 0.43              | 1.34   |
| dragon  | 1   | -      | 2.27              | 1.58              | 1.43   |
| random  | 7   | -      | 346               | 182               | 1.95   |
| fractal | 2   | -      | 6.75              | 4.1               | 1.64   |

The speed-up is not negligible. The unusual `random` timings is because we don't compute enclosing radius anymore.

BTW, I needed to update locally this [line](https://github.com/giotto-ai/giotto-ph/blob/7b07c3cfe3f336a5d74891b6e5cc2ce52de4fe0a/gph/python/ripser_interface.py#L583):
From:
```python
elif not compute_enclosing_radius:
```
To
```python
elif not compute_enclosing_radius and thresh != np.inf:
```

Because otherwise it converted the dense distance matrix to a sparse distance matrix using threshold `np.inf`.

2 questions @ulupo :

1. How to apply a threshold after computing weights ? Could we encounter any issue with the current logic ?
2. Could the enclosing radius still be applied when using `weights` when no threshold is set ?

gph/python/ripser_interface.py
```diff
@@ -558,30 +558,27 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
         if weights is not None:
             dm = _compute_weights(dm, weights, weight_params, n_points)
 
-        compute_enclosing_radius = False
+        apply_user_threshold = thresh != np.inf
         nonzero_in_diag = (dm.diagonal() != 0).any()
-        if not nonzero_in_diag:
+        if not nonzero_in_diag and not apply_user_threshold:
```
MonkeyBreaker, line 563: @ulupo I prepare a suggestion in the logic for when to apply the threshold.
Let me know what you think of this and if the variables names make sense to you

gph/src/ripser.h
```diff
@@ -280,24 +280,30 @@ struct compressed_distance_matrix {
...
+            for (size_t j = 0; j <= i; ++j)
                 rows[i][j] = mat(i, j);
     }
 
-    value_t operator()(const index_t i, const index_t j) const;
+    inline value_t mat_tri_val(const index_t min, const index_t max) const;
+
+    value_t operator()(const index_t i, const index_t j) const
+    {
+        const auto& p = std::minmax(i, j);
```
MonkeyBreaker, line 304: This is a bit tricky, this currently computes the min and max values, due to compiler optimization, no variable is created. It just helps to read the code.

Something interesting though, is that `std::minmax` with `SS2` optimization enable, allows to perform a branchless computation. In our case, it doesn't provide any performance enhancement/slowdown, but the code is ready that with specific compiler optimization enable. It could provide further speed-up.

gph/src/ripser.h
```diff
@@ -280,24 +280,30 @@ struct compressed_distance_matrix {
...
+        return mat_tri_val(p.first, p.second);
```
MonkeyBreaker, line 305: Unfortunately, this we need this syntax because our compiler minimum standard supported version is `c++14`. If the minimal supported version would be `c++17`, the we could use the following syntax:

```cpp
if constexpr(std::is_same_v<Layout, compressed_lower_distance_matrix>)
    return rows[p.second][p.first];
else
    return rows[p.first][p.second];
```

This is IMO, much cleaner code, but I wouldn't make `c++17` the minimum requirement just for this feature. This feature is also interesting, because it is evaluated at compile time, that should produce similar code to the current implementation

ulupo:
@MonkeyBreaker, thanks for all the work!  I must admit that I am struggling a bit to understand this PR even though I was the one to propose it, oops 👎 

I think it would help me and readers of this thread if the PR description included some background as follows:

- (briefly) how the various input cases are dispatched to the structs and backends we have inherited from Ripser(.py);
- why we think the original authors of `ripser.py` opted to only support diagonal entries via the sparse computer;
- at high level, how this PR manages to do what the authors of `ripser.py` could not/did not want.

Sorry to give you more work!

MonkeyBreaker:
> Thanks for all the work! I must admit that I am struggling a bit to understand this PR even though I was the one to propose it, oops 👎🏼

No worries, it is maybe easy to me to understand, because I am the one who made it 😆 

> (briefly) how the various input cases are currently dispatched to the structs and backends we have inherited from Ripser(.py)

From the standard datasets we used, I do not include the ones where a threshold is explicitly set (o3_1024, o3_4096 and torus), because when a threshold is applied we automatically switch to a sparse representation.

From the remaining datasets, this PR **only** impacts when non zero values are present in the distance matrix, e.g. using `weights`. The following tables shows the data structure used when non zeros values were present in the distance matrix:

| dataset  | before | now  | 
|---------|-------------------|-------------------|
| sphere  | sparse         | dense   |
| dragon  | sparse              | dense  |
| random  | sparse               | dense   |
| fractal |  sparse               | dense   |

> why we think the original authors of ripser.py opted to only support diagonal entries via the sparse computer;

Because in the original implementation of the C++ backend (`ripser`), the dense representation explicitly returned 0 for all diagonals values, see this [line](https://github.com/scikit-tda/ripser.py/blob/8cb3e012dbb6315c8d75d278862b64aa88df562e/ripser/ripser.cpp#L317).

Now in our case, we store the diagonals, even when their value is 0, but this allows us at the same time to support non-zero diagonals values.

About the sparse representation, it was modified to store in `vertex_births` the diagonals values. I am not sure why they did not store it directly into `neighbors` matrix, but anyway it is maybe cleaner this way.

Why they modified only the sparse representation, I don't know.

> at high level, how this PR manages to do what the authors of ripser.py could not/did not want.

By in the C++ backend, to add support to non zero values in the diagonals for the `dense` matrix representation.
In the Python part, an important change is this:

```python
DParam = dm[np.triu_indices(dm.shape[0])].astype(np.float32)
```

It allows to have a compressed representation of a full distance matrix, but in also includes the diagonal.

@ulupo I hope and provide more understanding on what this PR does, otherwise let me know if I must provide more details somewhere.

MonkeyBreaker:
@ulupo I made the changes we discussed yesterday.
Even with the new changes, the speed-up is quite similar to the one I published 20 days ago !

Let me know if this is more in the direction you had in mind.

ulupo:
@MonkeyBreaker looks good!  The C++ code is, I think, simpler this way.  I guess some of the complexity has been shifted to the bindings.

It would be good to have a small test with a sparse matrix with non-zero values in the diagonal and its dense version lead to the same result.  I can do this if you don't have the time.

gph/python/test/test_ripser.py
```diff
@@ -431,3 +431,24 @@ def test_unsupported_coefficient():
...
+
+
+@settings(deadline=500)
+@given(dm_dense=get_dense_distance_matrices())
+def test_non_0_diagonal_internal_representation(dm_dense):
+    """Checks that when passing a full distance matrix with non zero values
+    in the diagonal in dense or sparse format. Produces the exact same results
+    """
+
+    diagonal = np.random.random(dm_dense.shape[0])
```
ulupo, line 443:
```suggestion
    """Checks that, when passing a full distance matrix with non-zero values in
    the diagonal, the result is the same regardless of whether the input is in
    dense or sparse format."""
    diagonal = np.random.random(dm_dense.shape[0])
```

ulupo (changes requested):
Just a small wording suggestion

ulupo (commented):
I feel like `nonzero_in_diag` should not even exist any more in `ripser_interface` after this PR.  Is it still needed?

MonkeyBreaker:
> I feel like nonzero_in_diag should not even exist any more in ripser_interface.py after this PR. Is it still needed?

I still use it to not compute enclosing radius if non-zero values are present in the diagonal.
I don't know if we should compute the enclosing radius if non-zero values in the diagonal are present.

ulupo:
Thanks for the reply!

> I don't know if we should compute the enclosing radius if non-zero values in the diagonal are present.

Yes, in the dense case and if no threshold is passed we should compute and use the enclosing radius in all cases

MonkeyBreaker:
> Yes, in the dense case and if no threshold is passed we should compute and use the enclosing radius in all cases

Well, proceeding to this started to make one fail from time to time.
The test is `test_collapser_with_negative_weights` in `test_ripser.py`test file.

Attached you will find one distance matrix that makes fail this test, to make it easier to reproduce I wrote the following script.

```python
#!/usr/bin/env python

import numpy as np
from gph.python import ripser_parallel
from scipy.sparse import coo_matrix
from numpy.testing import assert_almost_equal
import sys

dm = np.loadtxt(sys.argv[1], dtype=np.float64)
[failed_dense.txt](https://github.com/giotto-ai/giotto-ph/files/8010247/failed_dense.txt)


a = ripser_parallel(dm, metric='precomputed', maxdim=2)['dgms']
b = ripser_parallel(coo_matrix(dm), metric='precomputed', maxdim=2,
                    collapse_edges=True)['dgms']

for dim, (dgms1, dgms2) in enumerate(zip(a, b)):
    print('dim: {}'.format(dim))
    print(dgms1)
    print(dgms2)
    assert_almost_equal(dgms1, dgms2)

```

It should fail when comparing results in dimension 2.

The only change I made to enable computing enclosing radius with non-zero diagonal was as follow:
```python
        if not nonzero_in_diag and not apply_user_threshold:
```
to
```python
        if not apply_user_threshold:
```

If you look at the input distance matrix, it contains negative edges, I don't know if this could be an ill formed matrix and therefore the current fail is irrelevant.

@ulupo I know that you are away for a moment, no need to rush this PR, let come back to it again after you are back :)

ulupo:
Thanks @MonkeyBreaker for the thorough testing!

Looking at that test, the matrix that is constructed there is not symmetric.  I think this means that the computed enclosing radius will not be correct, making the test fail.  This was not observed before because the matrix was converted to sparse format and the enclosing radius was not computed.

MonkeyBreaker:
@ulupo great, thank you for spotting this obvious mistake ! I should add to myself this kind of check to verify that data provided is good ... Not the first time I encounter this issue.

I pushed the changed you suggested slightly different. I also update the interface and below you will find the benchmark with now integrating the enclosing radius 👍🏼 . Looking at the results, I think it was worth it :)

| dataset | dim | thresh | before timing [s] | after timing  [s] | factor |
|---------|-----|--------|-------------------|-------------------|--------|
| sphere  | 2   | -      | 0.58              | 0.39              | 1.49   |
| dragon  | 1   | -      | 2.27              | 1.32              | 1.72   |
| random  | 7   | -      | 346               | 2.26              | 153    |
| fractal | 2   | -      | 6.75              | 3.42              | 1.97   |

gph/python/ripser_interface.py
```diff
@@ -558,30 +556,27 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
...
-        if nonzero_in_diag:
-            # Convert to sparse format, because currently that's the only one
-            # handling nonzero births
-            (row, col) = np.triu_indices_from(dm)
-            data = dm[(row, col)]
-            if collapse_edges:
+            thresh = _ideal_thresh(dm, thresh)
+
+        if collapse_edges:
+            if (dm.diagonal() != 0).any():
```
ulupo, line 569: Can we not simplify (remove?) this logic now that we do support non-zero births?  At least it seems to me that the comment is now outdated.

gph/python/test/test_ripser.py
```diff
@@ -452,4 +475,4 @@ def test_infinite_deaths_always_essential():
     gens_fin_dim1 = gens[1][1]
 
     # With this example no finite generators in dimension 1 shall be found
-    assert len(gens_fin_dim1) == 0
+    assert len(gens_fin_dim1) == 0
```
ulupo, line 478: Missing newline at the end of the file.

ulupo (changes requested):

gph/python/ripser_interface.py
```diff
@@ -605,9 +600,10 @@ def ripser_parallel(X, maxdim=1, thresh=np.inf, coeff=2, metric="euclidean",
             )
     else:
         # Only consider strict upper diagonal
```
ulupo, line 602:
```suggestion
        # Only consider upper diagonal
```
("strict" would mean no diagonal)

gph/python/ripser_interface.py
ulupo: To be more precise/constructive here, my point is that it may be cleaner to create a `_collapse_dense` wrapper function more or less as follows:

```
def _collapse_dense(dm, thresh):
    # Diagonal data (in case it is nonzero)
    data_diag = dm.diagonal()
    row_diag = np.arange(n_points)
    col_diag = row_diag.copy()
    # Off-diagonal collapsed edges
    row, col, data = gph_collapser.\
        flag_complex_collapse_edges_dense(dm, thresh)

    return (np.hstack([row_diag, row]),
            np.hstack([col_diag, col]),
            np.hstack([data_diag, data]))
```
Then one could call ``row, col, data = _collapse_dense(dm.astype(np.float32), thresh)`` without conditionals.

Btw, why don't we convert to `np.float32` in the case of `_collapsed_coo`?

MonkeyBreaker: Umbe, you discovered a bug ...
In collapser, the diagonal was always discarded, dense and sparse case.
I will push a commit here that you can have a look and let me know what do you think.

The commit fixes the bindings to include the diagonal. In the if condition for dense and sparse case, you will see 2 changes:
1. Include diagonal
2. Check if the value in the entry is different than 0. (To prevent including 0 entries from a 0-diagonal per example)

I also made slightly different function as you suggested, let me know what do you think.

MonkeyBreaker: > Btw, why don't we convert to np.float32 in the case of _collapse_coo?

I think I was just inconsistent ...

MonkeyBreaker: Wait, there was no bug, it was by purpose that we did not include the diagonal in collapser.
Sorry for that, but anyway, I will revert just the part related to collapser bindings.

gph/python/test/test_ripser.py
ulupo, line 478:
```suggestion
    assert len(gens_fin_dim1) == 0

```

gph/python/ripser_interface.py
ulupo: Thanks @MonkeyBreaker!  I agree that there was no bug.

> I also made slightly different function as you suggested, let me know what do you think.

Thanks!  I still think that this logic deserves to be encapsulated into a function as I initially suggested.  After all, we do something logically very similar to `_collapse_coo`, so it would be more consistent.  Furthermore, I still slightly favour the original code I suggested over the use of `diags`, because using `diags` is slower (about 300 times slower in my machine!) as it performs several checks and format conversions internally.  It's true that using `diags` lets us skip some nonzero values if there are any, but the memory impact of storing the entire diagonal is negligible anyway.

> I think I was just inconsistent ...

Which way should we go to be consistent?  Since the collapser bindings use 64-bit, I am tempted to say that we should probably delay conversion to 32-bit to later, so we can use the extra precision if necessary until we really *have* to downcast inside ripser.

gph/src/ripser.h
ulupo: What do you think about adding a short version of this as an inline comment?  I fear that otherwise some of this wisdom will get lost!

gph/src/ripser.h
ulupo: Same suggestion about an inline comment as above.

gph/python/ripser_interface.py
MonkeyBreaker: > Thanks! I still think that this logic deserves to be encapsulated into a function as I initially suggested. After all, we do something logically very similar to _collapse_coo, so it would be more consistent.

Sure !

>  still slightly favour the original code I suggested over the use of diags, because using diags is slower (about 300 times slower in my machine!) 

I wanted to test this, I was a bit surprised about the 300 magnitude order. The next Figure shows a comparison between both method when increasing the size of the diagonal (I was a bit bored tonight ...):
image

The Numpy approach is always faster, but I think that even for a diagonal of 300000 points, it takes less than 10ms. I would not choose the Numpy approach over the Scipy looking at the performances. But because you mention that diags skips 0 values, that is not good for us. We need the entire diagonal, otherwise we would shift elements from diagonal, which not desired.

I will go for the Numpy solution, let me know what do you think and update if necessary ;)

gph/src/ripser.h
MonkeyBreaker: Done for both !
Feel free to modify and integrate if necessary 👍🏼 
You can then resolve conversations

gph/python/ripser_interface.py
MonkeyBreaker: > Which way should we go to be consistent? Since the collapser bindings use 64-bit, I am tempted to say that we should probably delay conversion to 32-bit to later, so we can use the extra precision if necessary until we really have to downcast inside ripser.

Are you sure about collapser using 64-bit ? Looking at the bindings, the Filtration_value is a simple float (32-bit), in C++ double is 64-bit.

ulupo: > Are you sure about collapser using 64-bit ?
I'm not, no.  I was only fairly confident that GUDHI tend to use 64-bit precision for persistent homology, so expected the same of their collapser and hence also ours.

MonkeyBreaker: > I'm not, no. I was only fairly confident that GUDHI tend to use 64-bit precision for persistent homology, so expected the same of their collapser and hence also ours.

How, in our case, I use 32-bit because also ripser uses 32-bit, I did not see the point of computing one operation in 32-bit and the other in 64-bit. This shall be addressed in a future PR for support for 64-bit precision ;) (Hopefully soon, if I have some time)

ulupo: I was silly when I talked about 300x speedup, but I do confirm an 80x speedup in the following comparison:
```
diags(dm.diagonal(), format="coo")
```
vs
```
diag_data = dm.diagonal()
row = np.arange(100)
col = row.copy()
```
for dm a 100 x 100 matrix.  The first test takes 40 mus while the second takes 500 ns.

MonkeyBreaker: I see why. In my opinion with "small" arrays, we maybe pay the price of all the checks done inside of `diags`.
I was just surprised about the difference 😄 .
Thank you for pointing out performances and also the behavior of `diags`, discarding zeros in our case is not the intended behavior.

gph/python/ripser_interface.py
```diff
@@ -106,6 +107,21 @@ def _collapse_coo(row, col, data, thresh):
             np.hstack([data_diag, data]))
 
 
+def _collapse_dense(dm, thresh):
+    """Run edge collapser on off-diagonal data and then reinsert diagonal
+    data if any non zero values is present"""
```
ulupo, line 112:
```suggestion
    data if any non-zero value is present"""
```

gph/python/ripser_interface.py
```diff
@@ -108,6 +107,21 @@ def _collapse_coo(row, col, data, thresh):
...
```
ulupo, line 112:
```suggestion
    data if any non-zero value is present."""
```

gph/src/ripser.h
ulupo: Awesome, thank you!

gph/python/ripser_interface.py
```diff
@@ -106,6 +107,21 @@ def _collapse_coo(row, col, data, thresh):
...
+    row, col, data = gph_collapser.flag_complex_collapse_edges_dense(
+        dm.astype(np.float32), thresh)
+
+    if (dm.diagonal() != 0).any():
+        indices = np.arange(dm.diagonal().shape[0])
+        row = np.hstack([indices, row])
+        col = np.hstack([indices, col])
+        data = np.hstack([dm.diagonal(), data])
```
ulupo, line 120:
```suggestion
    data_diag = dm.diagonal() != 0
    if (data_diag != 0).any():
        indices = np.arange(data_diag.shape[0])
        row = np.hstack([indices, row])
        col = np.hstack([indices, col])
        data = np.hstack([data_diag, data])
```
to compute `dm.diagonal()` once instead of three times (and, to me at least, make the code slightly more readable).

gph/python/ripser_interface.py
```diff
@@ -106,6 +107,21 @@ def _collapse_coo(row, col, data, thresh):
             np.hstack([data_diag, data]))
 
 
+def _collapse_dense(dm, thresh):
+    """Run edge collapser on off-diagonal data and then reinsert diagonal
+    data if any non zero values is present"""
+    row, col, data = gph_collapser.flag_complex_collapse_edges_dense(
+        dm.astype(np.float32), thresh)
```
ulupo, line 114: I think we should downcast `dm` before passing it to `_collapse_dense`.  Currently, the final `data` might end up being 64-bit.

MonkeyBreaker: I am not sure I get what you mean.
You want to downcast just before calling `_collapse_dense`, i.e. `_collapse_dense(dm.astype(np.float32), thresh)` ?
Because as far as I understand, we do the same thing in `_collapse_coo` to downcast just before calling collapser.

IMO it is better to downcast inside of `_collapse_dense`, it keeps behavior encapsulated. And, at some point we will support both 32 and 64-bit precision for values in the distance matrix.

MonkeyBreaker:
The PR LGTM.
I think if we can resolve the last open conversation, we can proceed to merge 🎉

gph/python/ripser_interface.py
ulupo: I was thinking of calling `_collapse_dense(dm.astype(np.float32), thresh)`, but I would also be happy with 
```
def _collapse_dense(dm, thresh):
    """Run edge collapser on off-diagonal data and then reinsert diagonal
    data if any non-zero value is present."""
    dm = dm.astype(np.float32)
    row, col, data = \
        gph_collapser.flag_complex_collapse_edges_dense(dm, thresh)
```
The point is that I don't like the fact that in line 116 `data_diag` could be 64-bit.

MonkeyBreaker: Nice spotted !
You are right, I think that doing it inside of the function is better.
Because we hide some internals that I hope will be updated this year !
If you don't mind, I will proceed with my suggestion, I am happy if you can review it ;).

MonkeyBreaker: Changed pushed.
Repeated becnhmark, no regression from my latest one.

ulupo: Perfect!

ulupo (approved):
Yahoo!

Issue #56, opened 2021-11-23
Extend dense backend to cover case of non-zero diagonal weights


ulupo:
Using weighted Rips filtrations means a dense `dm` is computed but crucially it will typically have non-zero values along the diagonal.  Yet, we will use the sparse C++ backend in this case.

This causes a substantial loss in performance compared to what could be achieved with a dense backend, yet I think it is conceptually easy to cover such cases with the dense backend.  We just need to allow for the possibility to pass diagonal information as an additional piece of data, and use it in the 0-dimensional computation. The higher-dimensional computation should remain the same.

This could be a worthwhile target for v0.4.0 or later.

MonkeyBreaker:
Good idea !

I think that this will mostly impact `compressed_distance_matrix` class.

In the [initialization](https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/src/ripser.h#L301), we will need to include the diagonal which is currently skipped because it is assumed to be 0.

Also for the [operator](https://github.com/giotto-ai/giotto-ph/blob/2b3d6763af097762ca702222744ce097bb3588d7/gph/src/ripser.h#L298), this will remove a branch `i==j`.

This change will be interesting in terms of performance. 
I think that we will consume more memory in `compressed_distance_matrix`, but we will also remove a branch in `operator` and this function is called a lot.
We could maybe observe some surprise !

I think that the estimation for v0.4.0 is a good !

PR #62, opened 2022-02-06
Drop Python 3.6, enable Apple Silicon wheels for Python 3.8, 3.9 and 3.10, upgrade to manylinux2014

MonkeyBreaker:
This PR provide wheels for the new Apple Silicon for Python 3.8, 3.9 and 3.10.
Unfortunately, for the moment the wheels can be generated but cannot be tested because Github Action runner do not have a native ARM runner for MacOS.

I did not find if a public statement will provide in a near future support for native ARM runners. But on the meantime we can already generate the wheels and anybody that has a Mac with the new Silicon can perform the test on his/her machine and let us if it works. And only then, we will upload the new wheels on Pypi.

Unfortunately I do not have this kind of machine available.

Of course another option would be to emulate the hardware, but, for the moment I do not have the time to deploy this kind of feature.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -27,7 +27,7 @@ jobs:
           CIBW_TEST_COMMAND: python -m pytest {package}/gph/python/test
           CIBW_TEST_REQUIRES: pytest hypothesis
           # Should generate universal2 wheels for CP3.8 and CP3.9
-          CIBW_ARCHS_MACOS: x86_64 arm64 universal2
+          CIBW_ARCHS_MACOS: universal2
 
       # SciPy do not ship manylinux2010 wheels for Python 3.10, so we too build ours for manylinux2014
       - name: Build wheels for Python 3.10
@@ -39,7 +39,7 @@ jobs:
           CIBW_TEST_COMMAND: python -m pytest {package}/gph/python/test
           CIBW_TEST_REQUIRES: pytest hypothesis
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-          CIBW_ARCHS_MACOS: x86_64 arm64 universal2
+          CIBW_ARCHS_MACOS: universal2
 
       - uses: actions/upload-artifact@v2
         name: Upload wheels
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -20,25 +20,14 @@ jobs:
         env:
           # Specify which Python versions to build wheels
           # https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip
-          CIBW_BUILD: "cp37-* cp38-* cp39-*"
+          CIBW_BUILD: "cp37-* cp38-* cp39-*" "cp310-*" "cp311-*"
           # Skip 32 bit architectures, musllinux, and i686
           CIBW_SKIP: "*-win32 *-musllinux_x86_64 *_i686"
           CIBW_BEFORE_BUILD: python -m pip install cmake
           CIBW_TEST_COMMAND: python -m pytest {package}/gph/python/test
           CIBW_TEST_REQUIRES: pytest hypothesis
-          # Should generate universal2 wheels for CP3.8 and CP3.9
-          CIBW_ARCHS_MACOS: universal2
-
-      # SciPy do not ship manylinux2010 wheels for Python 3.10, so we too build ours for manylinux2014
-      - name: Build wheels for Python 3.10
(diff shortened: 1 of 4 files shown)


MonkeyBreaker:
The wheels are available [here](https://github.com/giotto-ai/giotto-ph/actions/runs/1802849116) If anybody with latest Apple Silicon can download them and test them locally I will be more than glad :)

ulupo:
> The wheels are available [here](https://github.com/giotto-ai/giotto-ph/actions/runs/1802849116) If anybody with latest Apple Silicon can download them and test them locally I will be more than glad :)

I can do this!

ulupo:
@MonkeyBreaker I guess the wheels need to be re-generated since #56 was merged.  I'm ready to test them once you've done so.

MonkeyBreaker:
@ulupo [here](https://github.com/giotto-ai/giotto-ph/actions/runs/2024931315#artifacts) just out of the oven :)

ulupo:
I cannot even use the universal2 wheels on my machine. Opening the interpreter in the environment in which I have installed the wheels, I get:

```py
>>> import gph
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ulupo/miniforge3/envs/test-gph-py38/lib/python3.8/site-packages/gph/__init__.py", line 1, in <module>
    from .python import ripser_parallel
  File "/Users/ulupo/miniforge3/envs/test-gph-py38/lib/python3.8/site-packages/gph/python/__init__.py", line 1, in <module>
    from .ripser_interface import ripser_parallel
  File "/Users/ulupo/miniforge3/envs/test-gph-py38/lib/python3.8/site-packages/gph/python/ripser_interface.py", line 32, in <module>
    from ..modules import gph_ripser, gph_ripser_coeff, gph_collapser
ImportError: dlopen(/Users/ulupo/miniforge3/envs/test-gph-py38/lib/python3.8/site-packages/gph/modules/gph_ripser.cpython-38-darwin.so, 0x0002): tried: '/Users/ulupo/miniforge3/envs/test-gph-py38/lib/python3.8/site-packages/gph/modules/gph_ripser.cpython-38-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e'))
```

.github/workflows/wheels.yml
```diff
@@ -26,17 +26,20 @@ jobs:
...
-        uses: pypa/cibuildwheel@v2.2.2
+        uses: pypa/cibuildwheel@v2.3.1
         env:
           CIBW_BUILD: "cp310-*"
           CIBW_SKIP: "*-win32 *-musllinux_x86_64 *_i686"
           CIBW_BEFORE_BUILD: python -m pip install cmake
           CIBW_TEST_COMMAND: python -m pytest {package}/gph/python/test
           CIBW_TEST_REQUIRES: pytest hypothesis
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          CIBW_ARCHS_MACOS: x86_64 universal2
```
ulupo, line 42: Perhaps this should be `x86_64 arm64 universal2` to solve my issues?

ulupo: I added this, re-generated the wheels, but still get the same error as before!

Could this be related? https://github.com/Unidata/netcdf4-python/issues/1148#issuecomment-1017645753

ulupo:
I'm happy to help complete work on this PR.  As we discussed, Python 3.6 should be removed and we should adopt manylinux2014 for all python versions (now we only use it for Python 3.10).  Perhaps Python 3.7 can now be covered too with macOS ARM wheels?  I will then test the wheels on my M1 machine.

MonkeyBreaker:
Support to Python 3.6 dropped

.github/workflows/wheels.yml
```diff
@@ -16,27 +16,19 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.2.2
+        uses: pypa/cibuildwheel@v2.3.1
```
MonkeyBreaker, line 19: The latest release seems to be 2.9.0.
But I looking at the releases [notes](https://github.com/pypa/cibuildwheel/releases), it doesn't seem to me that a big bug was fixed.
I would prefer to update it only when we will also add support to Python 3.11, but not right yet.

What do you think @ulupo ?

ulupo: We only have a simple CI for now, and I'd prefer not to have surprisesn right now. So I agree with you!

MonkeyBreaker (commented):
LGTM

PR #66, opened 2022-08-16
Add macOS x86_64 wheels to cover Python 3.7 on Intel Macs

ulupo:


ulupo:
@MonkeyBreaker I've checked and this achieves the desired effect.  I have also included the x86_64 macOS wheels in the 0.2.2 release.

MonkeyBreaker:
@ulupo nice catch, sorry missed the 3.7 wheels on Mac, even if I told you I saw them ...
This should also be forwarded to pyflagser before merge of the latest PR about github actions

ulupo:
Ok great!  Should we merge this now and then proceed with `pyflagser`?

MonkeyBreaker:
> Ok great! Should we merge this now and then proceed with pyflagser?

Yes, definitely ! You can forward the change, otherwise I do it this evening.

PR #67, opened 2022-08-17
Include tests in build distributions

ulupo:
This is achieved in this PR by:
1. Adding `__init__.py` in the `tests` folder;
2. setting `CIBW_TEST_COMMAND: python -m pytest --pyargs gph` in cibuildwheel.


MonkeyBreaker:
Umbe if we do that, I think the wheels won't complete because the test won't be able to finish.
This is what I encountered on `pyflagser` earlier this week.

What is the reason to add that file ?

ulupo:
We are not distributing the test folder, I believe this is because of the missing __init__.  Users cannot test their wheels with this package, but they can with versions of `giotto-tda` and `pyflagser` on PyPI.

ulupo:
Just started a build, I'll have a look at how it fails.

MonkeyBreaker:
Yeah, but if you have wheels installed, the following case will fail I think:

```
pip install giotto-ph pytest`
git clone https://github.com/giotto-ai/giotto-ph
python -m pytest giotto-ph/python/test
```

The previous is steps are more or less what is done during the wheels generation.
I can try test this evening, but I am quite sure, wheels will fail to be generated.
About to export test in the wheels, I had a look in the past, but don't remember if I found a "fix".

ulupo:
I don't think we should worry about tests that fail if people have both the wheels and a clone installed and run the command from the root of a clone.  This would be a bad thing to do anyway.

ulupo:
@MonkeyBreaker using `pytest --pyargs` as we do in the giotto-tda Azure pipeline seems to work in the sense that it does not break the wheel building and testing GitHub Action. The artifacts are correctly produced and the test folders are inside the distributed  packages.  See https://docs.pytest.org/en/7.1.x/how-to/usage.html.  I think cibuildwheel would prefer us to do things differently (as you initially set them up).

I'll test whether `pytest gph` works with an installation from wheels (from any directory, and in the absence of a clone of the repo).

MonkeyBreaker:
@ulupo nice !
What a good surprise that they was an option for that.

Just confirmed that on the new wheels the test are present:

``` 
creating: gph/python/test/
  inflating: gph/python/ripser_interface.py
  inflating: gph/python/__init__.py
  inflating: gph/python/test/test_ripser.py
  inflating: gph/python/test/test_wrp.py
  inflating: gph/python/test/__init__.py
  inflating: gph/python/test/test_collapser.py
```

I will proceed this evening with the changes on pyflagser then, great news !
Also, we can be have a better feature parity for giotto-tda using github actions hopefully soon !

MonkeyBreaker (approved):

ulupo:
> I'll test whether pytest gph works with an installation from wheels (from any directory, and in the absence of a clone of the repo).

In fact, `pytest gph` does not work.  It throws an error:
```
ERROR: file or directory not found: gtda
```

But `pytest --pyargs gph` (same as in the CI) works for installations from PyPI!  I think `pytest` without the `pyargs` argument is just checking for tests in the directories available from the execution path.

ulupo:
@MonkeyBreaker just requesting your approval again since I added some info/questions after your approval!

MonkeyBreaker:
> file or directory not found: gtda

Wait when running gph you have an error about gtda ? Did not get it right

> Currently, the giotto-tda documentation does not really explain this. It only explains how to run pytest for a developer installation.

Nobody complained so far, I don't know if we should improve anything here ?

> The question is: do we want to ship these tests with the wheels (and thus use pytest --pyargs in the CI and maybe also tell the users they can run tests this way) in all our projects, or in none of them?

To be honest, I think that users won't care about it. Who might more care are developers, but, in developers case they should have cloned the repository. I personally don't have a strong opinion in favor of having the test in the wheels, but neither do I have a strong opinion about not having them. I let you decide ;)

ulupo:
> Wait when running gph you have an error about gtda ? Did not get it right

Sorry, this was a typo, I meant to say `gph`!  Exactly the same happens with `giotto-tda` and `gtda`.

> To be honest, I think that users won't care about it. Who might more care are developers, but, in developers case they should have cloned the repository. I personally don't have a strong opinion in favor of having the test in the wheels, but neither do I have a strong opinion about not having them. I let you decide ;)

Ok!  I don't see a big disadvantage in including tests in the wheels now that the CI works, so I am very slightly in favour of packaging things in the same way as giotto-tda and pyflagser.

PR #71, opened 2023-07-14
pybind11 module updated

rballeba:
This update allows to install giotto-ph on Python 3.11 and below. The project only needed to update the pybind11 library.


rballeba:
@ulupo This is the pull request for pybinds version in giotto-ph.

ulupo (approved):
Excellent!  I've also updated `wheels.yml` to add Python 3.11 there, so that 3.11 wheels can also be built. Will merge if CI passes.

